PR #38, opened 2018-05-03
Replace org.json usage with Jackson

### LINKED ISSUE #29: Avoid usage of org.json library
The driver currently has a dependency on `org.json`

```xml
 <dependency>
      <groupId>org.json</groupId>
      <artifactId>json</artifactId>
      <version>20140107</version>
</dependency>
```

This library license is marked as [Category-X][1] and hence such a library [cannot be used as a dependency in Apache Project][2]. Would it be possible to replace this library with some other json library? Otherwise it would become tricky to use azure-cosmosdb-java for Apache projects

[1]: https://www.apache.org/legal/resolved.html#category-x
[2]: https://www.apache.org/legal/resolved.html#optional

chetanmeh:
This PR is an initial attempt to fix #29 as thats the only issue blocking [CosmosDB support in OpenWhisk](apache/incubator-openwhisk#3562)

As changes are across multiple files this PR only touches those parts which are required for the switch. To keep the changes at minimum certain possible optimizations (due to Jackson features) have yet not been implemented and can be done in a later PR

1. Use [JsonNode.at(path)][1] to get node at path in `JsonSerializable#getObjectByPath`
2. Optimize construction of POJO instances - Currently `JsonSerializable` serializes existing json instance to string and then construct the pojo instance. Possibly this can be replaced with [deepCopy][2]

#### Behavior change

1. Use of single quotes - Some tests were using single quotes in json strings. They have been modified to use double quotes
2. `JsonSerializable` now uses `ObjectNode` instead of `JSONObject` from `org.json`

#### Test Results

On my fork the test mostly pass. Seeing intermittent failure in (such failures also seen on master)

1. `BackPressureCrossPartitionTest.query`
2. `ChangeFeedTest.changesFromParitionKeyRangeId_FromBeginning`

@moderakh Please have a look

[1]: https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/JsonNode.html#at-java.lang.String-
[2]: https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/JsonNode.html#deepCopy--

### DIFF
--- a/examples/src/test/java/com/microsoft/azure/cosmosdb/rx/examples/DocumentCRUDAsyncAPITest.java
+++ b/examples/src/test/java/com/microsoft/azure/cosmosdb/rx/examples/DocumentCRUDAsyncAPITest.java
@@ -135,7 +135,7 @@ public class DocumentCRUDAsyncAPITest {
      */
     @Test
     public void testCreateDocument_Async() throws Exception {
-        Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document doc = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         Observable<ResourceResponse<Document>> createDocumentObservable = asyncClient
                 .createDocument(getCollectionLink(), doc, null, true);
 
@@ -160,7 +160,7 @@ public class DocumentCRUDAsyncAPITest {
      */
     @Test
     public void testCreateDocument_Async_withoutLambda() throws Exception {
-        Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document doc = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         Observable<ResourceResponse<Document>> createDocumentObservable = asyncClient
                 .createDocument(getCollectionLink(), doc, null, true);
 
@@ -197,7 +197,7 @@ public class DocumentCRUDAsyncAPITest {
      */
     @Test
     public void testCreateDocument_toBlocking() throws DocumentClientException {
-        Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document doc = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         Observable<ResourceResponse<Document>> createDocumentObservable = asyncClient
                 .createDocument(getCollectionLink(), doc, null, true);
 
@@ -249,7 +249,7 @@ public class DocumentCRUDAsyncAPITest {
         // Create 10 documents
         List<Observable<ResourceResponse<Document>>> listOfCreateDocumentObservables = new ArrayList<>();
         for (int i = 0; i < 10; i++) {
-            Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", i, i));
+            Document doc = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", i, i));
 
             Observable<ResourceResponse<Document>> createDocumentObservable = asyncClient
                     .createDocument(getCollectionLink(), doc, null, false);
@@ -288,7 +288,7 @@ public class DocumentCRUDAsyncAPITest {
      */
     @Test
     public void testCreateDocument_toBlocking_DocumentAlreadyExists_Fails() throws DocumentClientException {
-        Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document doc = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         asyncClient.createDocument(getCollectionLink(), doc, null, false).toBlocking().single();
 
         // Create the document
@@ -314,7 +314,7 @@ public class DocumentCRUDAsyncAPITest {
      */
     @Test
     public void testCreateDocument_Async_DocumentAlreadyExists_Fails() throws Exception {
-        Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document doc = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         asyncClient.createDocument(getCollectionLink(), doc, null, false).toBlocking().single();
 
         // Create the document
@@ -341,13 +341,13 @@ public class DocumentCRUDAsyncAPITest {
     @Test
     public void testDocumentReplace_Async() throws Exception {
         // Create a document
-        Document createdDocument = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document createdDocument = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         createdDocument = asyncClient.createDocument(getCollectionLink(), createdDocument, null, false).toBlocking()
                 .single().getResource();
 
         // Try to replace the existing document
         Document replacingDocument = new Document(
-                String.format("{ 'id': 'doc%d', 'counter': '%d', 'new-prop' : '2'}", 1, 1));
+                String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\", \"new-prop\" : \"2\"}", 1, 1));
         Observable<ResourceResponse<Document>> replaceDocumentObservable = asyncClient
                 .replaceDocument(getDocumentLink(createdDocument), replacingDocument, null);
 
@@ -370,12 +370,12 @@ public class DocumentCRUDAsyncAPITest {
     @Test
     public void testDocumentUpsert_Async() throws Exception {
         // Create a document
-        Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document doc = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         asyncClient.createDocument(getCollectionLink(), doc, null, false).toBlocking().single();
 
         // Upsert the existing document
         Document upsertingDocument = new Document(
-                String.format("{ 'id': 'doc%d', 'counter': '%d', 'new-prop' : '2'}", 1, 1));
+                String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\", \"new-prop\" : \"2\"}", 1, 1));
         Observable<ResourceResponse<Document>> upsertDocumentObservable = asyncClient
                 .upsertDocument(getCollectionLink(), upsertingDocument, null, false);
 
@@ -398,7 +398,7 @@ public class DocumentCRUDAsyncAPITest {
     @Test
     public void testDocumentDelete_Async() throws Exception {
         // Create a document
-        Document createdDocument = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
+        Document createdDocument = new Document(String.format("{ \"id\": \"doc%d\", \"counter\": \"%d\"}", 1, 1));
         createdDocument = asyncClient.createDocument(getCollectionLink(), createdDocument, null, false).toBlocking()
                 .single().getResource();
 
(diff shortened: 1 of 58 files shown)


moderakh:
@chetanmeh thanks for the PR. As there are breaking changes in the PR, our team is discussing the way forward with the breaking changes. We will get back to you.

examples/src/test/java/com/microsoft/azure/cosmosdb/rx/examples/DocumentCRUDAsyncAPITest.java
```diff
@@ -135,7 +135,7 @@ public void shutdown() throws DocumentClientException {
      */
     @Test
     public void testCreateDocument_Async() throws Exception {
-        Document doc = new Document(String.format("{ 'id': 'doc%d', 'counter': '%d'}", 1, 1));
```
srinathnarayanan: Can you add OBJECT_MAPPER.configure(JsonParser.Feature.ALLOW_SINGLE_QUOTES, true) and OBJECT_MAPPER.configure(JsonParser.Feature.ALLOW_TRAILING_COMMA, true) to the JsonSerializable class to have a relazed json representation in the strings?

chetanmeh: That can be done but I would prefer that test prepare a proper json as that flag looks like a work around and to be used only for compatibility purposes

chetanmeh:
@moderakh I understand change is bigger here. So take your time and let me know if any aspect needs to be improved

examples/src/test/java/com/microsoft/azure/cosmosdb/rx/examples/DocumentCRUDAsyncAPITest.java
srinathnarayanan: This would be a large breaking change for existing customers, since we already have a version of the async SDK out that supports single quotes and trailing commas. This is also the default behavior of the sync SDK. Can you also revert the relevant changes to the tests? Thanks!

sdk/src/main/java/com/microsoft/azure/cosmosdb/JsonSerializable.java
```diff
@@ -81,68 +78,25 @@ protected JsonSerializable(String jsonString, ObjectMapper objectMapper) {
...
-            int length = Array.getLength(array);
-            for (int i = 0; i < length; i++) {
-                ar.add(Array.get(array, i));
-            }
-            return ar.toArray();
-        } else {
-            return (Object[]) array;
-        }
+    private static HashMap<String, Object> toMap(ObjectNode object) {
+       return OBJECT_MAPPER.convertValue(object, HashMap.class);
```
srinathnarayanan: replace with this.getMapper() so that it uses OBJECT_MAPPER (default) or this.om (user's object mapper)

sdk/src/main/java/com/microsoft/azure/cosmosdb/internal/Utils.java
```diff
@@ -369,6 +373,37 @@ public static ObjectMapper getSimpleObjectMapper() {
         return Utils.simpleObjectMapper;
     }
 
+    public static ObjectNode newObjectNode() {
+        return Utils.simpleObjectMapper.createObjectNode();
+    }
+
+    public static JsonNode fromJson(String json){
```
srinathnarayanan: Can you move the fromJson, toJson, toPrettyJson methods out of Utils to the actual classes that call them? This was we can pass more verbose and meaningful messages. For example, in JsonSerializable and PartitionKey, we can raise an illegalArgumentException when parsing of the string fails.

sdk/src/main/java/com/microsoft/azure/cosmosdb/JsonSerializable.java
```diff
@@ -230,45 +173,35 @@ public void remove(String propertyName) {
             this.propertyBag.put(propertyName, castedValue != null ? castedValue.propertyBag : null);
         } else {
             // POJO
-            try {
-                this.propertyBag.put(propertyName, new JSONObject(this.getMapper().writeValueAsString(value)));
-            } catch (IOException e) {
-                throw new IllegalArgumentException("Can't serialize the object into the json string", e);
-            }
+            this.propertyBag.set(propertyName, OBJECT_MAPPER.valueToTree(value));
```
srinathnarayanan: use this.getMapper()

sdk/src/main/java/com/microsoft/azure/cosmosdb/internal/Utils.java
chetanmeh: Currently `fromJson` is being being used in 3 classes and `toJson` is being used in 5. So should I copy these methods to those classes and customize the error handling or should I just change exception here to `illegalArgumentException`

sdk/src/main/java/com/microsoft/azure/cosmosdb/JsonSerializable.java
chetanmeh: Ack. Done now

sdk/src/main/java/com/microsoft/azure/cosmosdb/JsonSerializable.java
chetanmeh: Ack. Done now

examples/src/test/java/com/microsoft/azure/cosmosdb/rx/examples/DocumentCRUDAsyncAPITest.java
chetanmeh: Ack. Given compatibility concern single quoted json needs to be supported. Would configure the mapper to do  that and revert changes in tests

sdk/src/main/java/com/microsoft/azure/cosmosdb/internal/Utils.java
srinathnarayanan: Move the methods and customize the error handling. You can use illegalArgumentException and illegalStateException as applicable (illegalArgumentException should be for API s on the surface level)

chetanmeh: Ack. Done now

sdk/src/main/java/com/microsoft/azure/cosmosdb/JsonSerializable.java
```diff
@@ -81,68 +81,25 @@ protected JsonSerializable(String jsonString, ObjectMapper objectMapper) {
...
-            List<Object> ar = new ArrayList<Object>();
-            int length = Array.getLength(array);
-            for (int i = 0; i < length; i++) {
-                ar.add(Array.get(array, i));
-            }
-            return ar.toArray();
-        } else {
-            return (Object[]) array;
-        }
+    private static HashMap<String, Object> toMap(ObjectNode object) {
```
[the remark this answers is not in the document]
chetanmeh: Missed that. Done now

sdk/src/main/java/com/microsoft/azure/cosmosdb/IncludedPath.java
```diff
@@ -106,14 +107,14 @@ public void setIndexes(Collection<Index> indexes) {
 
     private Collection<Index> getIndexCollection() {
         if (this.propertyBag != null && this.propertyBag.has(Constants.Properties.INDEXES)) {
-            JSONArray jsonArray = this.propertyBag.getJSONArray(Constants.Properties.INDEXES);
+            List<JsonNode> jsonArray = this.propertyBag.findValues(Constants.Properties.INDEXES);
```
srinathnarayanan: findValues() returns a List of ArrayNodes. hence the .asText() fails as the node is an arrayNode.

i.e, Instead of [{"kind":"Hash","dataType":"Number","precision":2}], it is returning [[{"kind":"Hash","dataType":"Number","precision":2}]]. 

change it to : ArrayNode jsonArray = (ArrayNode) this.propertyBag.get(Constants.Properties.INDEXES);

sdk/src/main/java/com/microsoft/azure/cosmosdb/PartitionKey.java
```diff
@@ -47,8 +49,7 @@
     @SuppressWarnings("serial")
     public PartitionKey(final Object key) {
         this.key = new Object[] {key};
-        JSONArray array = new JSONArray(this.key);
-        this.keyString = array.toString();
+        this.keyString = toJson(Utils.getSimpleObjectMapper().valueToTree(this.key));
         this.internalPartitionKey = PartitionKeyInternal.fromObjectArray(new ArrayList<Object>() {{ add(key); }}, true);
```
srinathnarayanan: we need to support ValueNodes being passed into PartitionKey.
for example : doc.set("field", "value"); new PartitionKey(doc.get("field"));
earlier : would call partitionKey constructor with String value
now : would call partitionKey with TextNode value
TextNode (or any ValueNodes) is not handled by PartitionKeyInternal.fromObjectArray
Hence I would suggest that you make JsonSerializable.getValue() a public static method (as it doesn't modify a member of JsonSerializable anyway), and modify this line as :  

add(key instanceof ValueNode ? JsonSerializable.getValue((ValueNode) key) : key);

srinathnarayanan: Or alternatively, push the check for ValueNode and subsequent deserialization into the PartitionKeyInternal.fromObjectArray method

sdk/src/main/java/com/microsoft/azure/cosmosdb/Permission.java
```diff
@@ -115,7 +115,7 @@ public PartitionKey getResourcePartitionKey() {
         PartitionKey key = null;
         Object value = super.get(Constants.Properties.RESOURCE_PARTITION_KEY);
         if (value != null) {
-            JSONArray arrayValue = (JSONArray) value;
+            ArrayNode arrayValue = (ArrayNode) value;
             key = new PartitionKey(arrayValue.get(0));
```
srinathnarayanan: This behavior is inconsistent with legacy behavior. We end up sending a ValueNode as parameter to PartitionKey's constructor. This isn't handled by PartitionKeyInternal.fromObjectArray. So instead, I would suggest that you make JsonSerializable.getValue as a public static method (since it isn't modifying any members of JsonSerializable as such) and use that to deserialize the value : 
i.e, modify the line to "  key = new PartitionKey(JsonSerializable.getValue(arrayValue.get(0))); "

chetanmeh: Missed the change in semantics. Fixed that now with 83a01b0 and added a test case to assert the behavior

sdk/src/main/java/com/microsoft/azure/cosmosdb/IncludedPath.java
chetanmeh: My bad ...  missed the change in semantics. Fixed that now with 9cefd0d  and added a test case to assert the behavior

sdk/src/main/java/com/microsoft/azure/cosmosdb/OfferV2.java
```diff
@@ -41,7 +42,7 @@
     public OfferV2(int offerThroughput) {
         this.setOfferVersion(Constants.Properties.OFFER_VERSION_V2);
         this.setOfferType("");
-        JSONObject content = new JSONObject();
+        ObjectNode content = Utils.newObjectNode();
```
srinathnarayanan: Can be replaced by Utils.simpleObjectMapper.createObjectNode(), and Utils.newObjectNode can be removed.

srinathnarayanan: Thanks for your quick updates to this PR. With this lat change, everything will look good. Can you Please change the base branch of the PR to dev instead of master ? I will merge it once you do. Thanks

srinathnarayanan:
I've changed the base branch to dev instead of master. There seem to be some merge conflicts. Once you resolve them, I'll merge this PR

sdk/src/main/java/com/microsoft/azure/cosmosdb/OfferV2.java
```diff
@@ -58,9 +59,9 @@ public OfferV2(Offer otherOffer) {
         this.setOfferVersion(Constants.Properties.OFFER_VERSION_V2);
         this.setOfferType("");
 
-        JSONObject content = this.getContent();
+        ObjectNode content = this.getContent();
         if (content == null) {
-            content = new JSONObject();
+            content = Utils.newObjectNode();
```
srinathnarayanan: Can be replaced with Utils.getSimpleObjectMapper().createObjectNode(), and Utils.newObject() can be removed

chetanmeh: Done now

chetanmeh:
Thanks @srinathnarayanan for all the review feedback. 

I have rebased this branch to dev (via `git rebase --onto origin/dev origin/master`) and also swithed to  Jackson API for some of the newly added classes like `UniqueKey` etc

moderakh:
Thanks @chetanmeh and @srinathnarayanan for completing this 👍

PR #41, opened 2018-05-05
Remove redundant sentences

otaviojava:
This PR just remove redundant sentence such as 
¨T extends Object¨ once all objects extend Java that is unnecessary and also a public method in the interface that is defined in the JVM spec.

https://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html#jls-9.4

### DIFF
--- a/sdk/src/main/java/com/microsoft/azure/cosmosdb/PermissionMode.java
+++ b/sdk/src/main/java/com/microsoft/azure/cosmosdb/PermissionMode.java
@@ -39,7 +39,7 @@ public enum PermissionMode {
 
     private int value;
 
-    private PermissionMode(int value) {
+    PermissionMode(int value) {
         this.value = value;
     }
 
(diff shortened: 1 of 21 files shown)


sdk/src/main/java/com/microsoft/azure/cosmosdb/PermissionMode.java
```diff
@@ -39,7 +39,7 @@
 
     private int value;
 
-    private PermissionMode(int value) {
+    PermissionMode(int value) {
```
moderakh: @otaviojava why are we dropping the private keyword?

moderakh: oh. this is enum, and enum constructor is private by default. ok makes sense.

moderakh:
Thanks for the PR @otaviojava 
looks good to me. merging it.

PR #43, opened 2018-05-09
Creates checkstyle Plugin

otaviojava:
That was the check style that I created, I used an Eclipse Foundation project as a template.
Do you think rules make sense?
I got these errors:
Can I got ahead and fix those?


otaviojava:
@moderakh what do you think?

otaviojava:
Done, could you give higher priority to this one? This PR changes a lot of classes.

moderakh:
@otaviojava 

Thanks for the PR and sorry for delayed update on this.

Having automated check-style is great, however this PR introduces a large number of breaking changes to the naming of public APIs which are already in use by end users.

PR #44, opened 2018-05-18
Adds support to Mapper in the SDK

otaviojava:
Adds support to a Mapper that's a wrapper that abstracts these operations to an entity.

E.g.: Given an Person entity:

```java
@Entity(databaseName = "AzureSampleFamilyDB")
public class Person implements Serializable {

    private String id;

    private String name;

    private Integer age;

}
```

You can create a Mapper unit that will handle the operations related to a particular class.

```java
        AsyncDocumentClient client = ...;
        MappingManager mappingManager = MappingManager.of(client);
        Mapper<Person> mapper = mappingManager.mapper(Person.class);
```

```java

        Person ada = new Person();
        ada.setId("ada_" + System.currentTimeMillis());
        ada.setName("Ada Lovelace");
        ada.setAge(20);

        Observable<Person> observable = mapper.save(ada);

        Observable<Person> findById = mapper.findById(ada.getId());


        FeedOptions queryOptions = new FeedOptions();
        queryOptions.setMaxItemCount(100);

        Observable<List<Person>> queryResult = mapper.query("SELECT * FROM Person WHERE Person.name = 'Ada Lovelace'", queryOptions);

         Observable<Void> observable = mapper.deleteById(ada.getId());

```

:warning: Both MappingManager and Mapper should have one instance at Application scope, that is easily handled with frameworks like Spring or CDI.

## Remain tasks:

- [ ] improve the documentation
- [ ] Write an article about this resource
- [ ] write a project sample
- [ ] Write a project sample and post using Spring framework and Mapper
- [ ] Write a project sample and post using JavaEE/JakartaEE  and Mapper

## Next steps to improve the Mapper

IMHO: This PR is just the first step to move forward an easy integration with Mapper.

1) The Repository interface is given an interface that has basic methods a Java developer just need to extend it, so the interface will already be implemented by the Mapper manager. 

```java
public interface PersonRepository extends Repository<Person> {
}
```
```java
    PersonRepository repository = mappingManager.mapper(PersonRepository.class);//ready to use
```

Done: check: https://github.com/Azure/azure-cosmosdb-java/pull/45
2) Support by query by annotation

```java

public interface PersonRepository extends Repository<Person> {
 @Query("SELECT * FROM Person WHERE Person.name = @name")
 Observable<List<Person>> usingQuery(@Param("name") String name);

}
```

Done: check: https://github.com/Azure/azure-cosmosdb-java/pull/45

[...]

PR #45, opened 2018-05-20
Adds support to Repository (A mapper who does a query dynamically)

otaviojava:
:bomb: This PR requires: https://github.com/Azure/azure-cosmosdb-java/pull/44

This PR adds supports to a Repository interface; this interface allows a Java developer creates a query without handle about its implementation.

Given this entity:

```java

@Entity(databaseName = "AzureSampleFamilyDB")
public class Person {

    private String id;

    private String name;

    private Integer age;

}

```

Going beyond the Mapper, a Java developer can create its who interface with queries.

```java
public interface PersonRepository extends Repository<Person> {

    @Query("SELECT * FROM Person WHERE Person.name = @name")
    Observable<List<Person>> findByName(@Param("name") String name);
}
```

```java
MappingManager manager = MappingManager.of(client);
PersonRepository repository = manager.repository(PersonRepository.class);
Observable<List<Person>> result = repository.findByName("Ada Lovelace");
```


moderakh:
@otaviojava
Thanks for the PR and sorry for the delayed update on this.

We decided to have support for object mapping outside of SDK as part of Spring project (https://github.com/Microsoft/spring-data-cosmosdb). So we have to reject the PR.

PR #48, opened 2018-05-29
vulnerable dependency

moderakh:


moderakh:
It was just for testing if snyc detects vulnerabilities in dependencies.

PR #52, opened 2018-06-19
Fix date format for non-english locale

fpaparoni:


moderakh:
@fpaparoni thanks for the PR.

I wonder if we can have a unit test for this which Utils.nowAsRFC1123() returns the correct value for Non US Locale? Can you please add that too?

fpaparoni:
Utils.nowAsRFC1123() now always returns a value for Locale.US because otherwise it isn't accepted by cosmosdb, so I don't understand a test for non US Locale.

fpaparoni:
Ok I understand the purpose of it and I pushed the unit test

sdk/src/test/java/com/microsoft/azure/cosmosdb/internal/UtilsTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class UtilsTest {
+    @Test(groups = { "internal" })
+    public void localeUS() {
+    	Locale.setDefault(Locale.ITALIAN);
+    	DateTimeFormatter RFC_1123_DATE_TIME = 
+    			DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss zzz",Locale.US);
+    	String time=Utils.nowAsRFC1123();
+    	Locale.setDefault(Locale.US);
+    	RFC_1123_DATE_TIME.parse(time);
```
moderakh: does RFC_1123_DATE_TIME.parse(time) fail without your bugfix?

fpaparoni: Yes because Utils.nowAsRFC1123 without bugfix returns time depending on Locale of your enviroment. Setting it to a Locale different from US and expecting to parse this String will throw an Exception of date format

moderakh:
Thank you @fpaparoni merging this PR.
This will be in the next release.

PR #61, opened 2018-08-27
Upgrade to newer release of the maven-javadoc-plugin

JonathanGiles:
Building the SDK under JDK 11 fails due to the outdated version of maven-javadoc-plugin specified in the Maven pom file. Version 2.9.1 was released in June 2013, and is not supported under JDK 11. I have changed this to instead use 3.0.1, which was released in May 2018. SDK compilation now succeeds under JDK 11 (as well as earlier JDK releases).


moderakh:
thanks @JonathanGiles

PR #96, opened 2019-04-30
Added support for Hash v2 

srinathnarayanan:
exposed the getter and setter for PartitionKeyVersion and the enum PartitionKeyDefinitionVersion


direct-impl/src/test/java/com/microsoft/azure/cosmosdb/PartitionKeyHashingTests.java
```diff
@@ -0,0 +1,94 @@
...
+
+import com.fasterxml.jackson.databind.node.NullNode;
+import com.microsoft.azure.cosmosdb.internal.routing.PartitionKeyInternalHelper;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+public class PartitionKeyHashingTests {
+
+    @Test
+    public void TestEffectivePartitionKeyHashV1() {
```
moderakh: to be consistent with the formatting of other tests please drop "Test" prefix or suffix from this method and other methods.

direct-impl/src/test/java/com/microsoft/azure/cosmosdb/PartitionKeyHashingTests.java
```diff
@@ -0,0 +1,94 @@
...
+import org.testng.annotations.Test;
+
+import com.fasterxml.jackson.databind.node.NullNode;
+import com.microsoft.azure.cosmosdb.internal.routing.PartitionKeyInternalHelper;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+public class PartitionKeyHashingTests {
+
+    @Test
```
moderakh: please add the "unit" test group annotation for all the unit tests here and in other test methods.

direct-impl/src/test/java/com/microsoft/azure/cosmosdb/PartitionKeyHashingTests.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    @Test
+    public void TestHashV2PartitionKeyDeserialization() {
+        String partitionKeyDefinitionStr = "{\"paths\":[\"/pk\"],\"kind\":\"Hash\",\"version\":2}";
+        PartitionKeyDefinition partitionKeyDef = new PartitionKeyDefinition(partitionKeyDefinitionStr);
+        assertThat(PartitionKeyDefinitionVersion.V2).isEqualTo(partitionKeyDef.getVersion());
+        assertThat(PartitionKind.Hash).isEqualTo(partitionKeyDef.getKind());
+        assertThat("/pk").isEqualTo(partitionKeyDef.getPaths().toArray()[0]);
```
moderakh: btw, I think for assertj and the other testing frameworks the actual value comes first and then you will have the expected value:
```java
assertThat(partitionKeyDef.getPaths()).isEqualTo("/pk");
```

here and for the other asserts.
the right ordering helps in generating better log if there is a failure.

direct-impl/src/test/java/com/microsoft/azure/cosmosdb/PartitionKeyHashingTests.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void TestHashV1PartitionKeyDeserialization() {
+        String partitionKeyDefinitionStr = "{\"paths\":[\"/pk\"],\"kind\":\"Hash\"}";
+        PartitionKeyDefinition partitionKeyDef = new PartitionKeyDefinition(partitionKeyDefinitionStr);
+        assertThat(partitionKeyDef.getVersion()).isNull();
+        assertThat(PartitionKind.Hash).isEqualTo(partitionKeyDef.getKind());
+        assertThat("/pk").isEqualTo(partitionKeyDef.getPaths().toArray()[0]);
```
moderakh: ditto.

christopheranderson:
We should also be able to take the Partition Key tests that have lengths greater than 100 bytes and make them use HashV2 instead to stop them from randomly failing.

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
```diff
@@ -76,15 +80,27 @@ public void setKind(PartitionKind kind) {
...
+                            : null;
+                }
+            }
         }
 
         return this.version;
     }
 
-    void setVersion(PartitionKeyDefinitionVersion version) {
+    public void setVersion(PartitionKeyDefinitionVersion version) {
```
moderakh: why do we want setVersion to be public? what's the use case?

srinathnarayanan: user can set version to V1 or V2, though default is V2 adter api version 2018-12-31

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
```diff
@@ -24,10 +24,14 @@
 package com.microsoft.azure.cosmosdb;
 
 import com.microsoft.azure.cosmosdb.internal.Constants;
+import com.microsoft.azure.cosmosdb.rx.internal.Strings;
 
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.WordUtils;
```
moderakh: this is deprecated.

Please use "org.apache.commons.text.WordUtils" instead.

srinathnarayanan: Already using the non deprecated version

christopheranderson:
Overriding because the CLA task appears to be unhealthy.

PR #107, opened 2019-05-08
Added support for accessing non partitioned collections using V3 SDK

srinathnarayanan:
- Exposed PartitionKey.None object which corresponds a PartitionKeyInternal object where components=null
- Added check for systemkey in PartitionKeyDefinition
- Bumped API version to 2018-12-31
- Added checks for null PartitionKeyInternal components in the places where it is accessed (DocumentQueryExecutionContextFactory and AddressResolver)
- Modified partition key extraction in case pk doesn't exists in the document. Earlier Undefined.Value was added, now the Value added depends on whether the pk is a system key or not.
- Removed partitionKey header modification in getHeaders as the partitionKey specified now (like PartitionKey.None) need not necessarily be serializable.
- Added test to create non partitioned collection with older API version using REST, and access it with SDK

Other changes:
- Added more files to gitignore
- Modified FeedResponseListValidator to check based on ids
- Added helper methods in CosmosTestSuiteBase


commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKey.java
```diff
@@ -34,7 +34,7 @@
 
     private PartitionKeyInternal internalPartitionKey;
 
-    PartitionKey(PartitionKeyInternal partitionKeyInternal) {
+    public PartitionKey(PartitionKeyInternal partitionKeyInternal) {
```
moderakh: PartitionKeyInternal is an internal construct, making this constructor public will expose the internal construct. Are we going to make PartitionKeyInternal public?

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
```diff
@@ -134,6 +136,27 @@ public void setPaths(List<String> paths) {
         this.paths = paths;
     }
 
+    /**
+     * Indicates if the partition key is generated by the system.
+     *
+     * @return the boolean indicating is it is a system key.
+     */
+    public Boolean isSystemKey() {
```
moderakh: why are we preferring Boolean over primitive boolean?

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
```diff
@@ -134,6 +136,27 @@ public void setPaths(List<String> paths) {
...
+        if (this.systemKey == null) {
+            this.systemKey = super.getBoolean(Constants.Properties.SYSTEM_KEY);
+        }
+
+        return this.systemKey;
+    }
+
+    public PartitionKeyInternal getNonePartitionKeyValue() {
```
moderakh: same as above PartitionKeyInternal is an internal construct. Are we planning to make this public?

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKey.java
moderakh: it is even in the internal namespace:
main/java/com/microsoft/azure/cosmosdb/internal/routing/PartitionKeyInternal.java

examples/src/main/java/com/microsoft/azure/cosmosdb/rx/examples/cosmosclient/BasicDemo.java
```diff
@@ -0,0 +1,220 @@
...
+import com.microsoft.azure.cosmos.CosmosContainerSettings;
+import com.microsoft.azure.cosmos.CosmosDatabase;
+import com.microsoft.azure.cosmos.CosmosItem;
+import com.microsoft.azure.cosmos.CosmosItemRequestOptions;
+import com.microsoft.azure.cosmos.CosmosItemResponse;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+
+public class BasicDemo {
```
moderakh: I think Bhaskar moved this file to another location. This is a duplicate file.

sdk/src/test/java/com/microsoft/azure/cosmos/CosmosPartitionKeyTests.java
```diff
@@ -0,0 +1,328 @@
...
+
+        responseStream = httpClient.submit(serverInfo, httpRequest).flatMap(clientResponse -> {
+            return toInputStream(clientResponse.getContent());
+        }).toBlocking().single();
+        String createdItemAsString = IOUtils.readLines(responseStream, "UTF-8").get(0);
+        assertThat(createdItemAsString).contains("\"id\":\"" + NON_PARTITIONED_CONTAINER_DOCUEMNT_ID + "\"");
+    }
+
+    @Test(groups = { "cosmosv3" }, timeOut = TIMEOUT)
+    public void testNonPartitionedCollectionOperations() throws Exception {
```
moderakh: remove "test" prefix and suffix from all the tests to be consistent with the existing tests.

moderakh:
``PartitionKeyInternal`` is an internal construct in com.microsoft.azure.cosmosdb.internal.routing package.

This PR is making it public. The existing public methods in ``PartitionKeyInternal`` are really meant to be used internally. for example
```java
   // PartitionKeyInternal
    public List<IPartitionKeyComponent> getComponents() {
        return this.components;
    }
```

examples/src/main/java/com/microsoft/azure/cosmosdb/rx/examples/cosmosclient/BasicDemo.java
srinathnarayanan: removed it

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKey.java
srinathnarayanan: I made it public since it is used by DocumentQueryExecutionContextFactory. 
I can instead make it use the public constructor that take Object, and have a check for Object being of type PartitionKeyInternal.

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
srinathnarayanan: This is to ensure that in case system_key does not exist in the Pkd, a null can be returned.

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
srinathnarayanan: This would only return a static object of PartitionKeyInternal. It can't be used in any of our APIs (once I make PartitionKey(PartitionKeyInternal) private).  No other implementation details are exposed. This is the .Net implementation too.

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
```diff
@@ -40,6 +41,7 @@
     private List<String> paths;
     private PartitionKind kind;
     private PartitionKeyDefinitionVersion version;
+    private Boolean systemKey;
```
pradeeptamma: In case of partitioned collections this tag may not be present in the returned JSON content. Can you clarify that the code works in such case.

sdk/src/test/java/com/microsoft/azure/cosmos/CosmosPartitionKeyTests.java
```diff
@@ -0,0 +1,328 @@
...
+
+    @Test(groups = { "cosmosv3" }, timeOut = TIMEOUT)
+    public void testMultiPartitionCollectionReadDocumentWithNoPk() throws InterruptedException {
+        String partitionedCollectionId = "PartitionedCollection" + UUID.randomUUID().toString();
+        String IdOfDocumentWithNoPk = UUID.randomUUID().toString();
+        CosmosDatabase createdDatabase = this.client.getDatabase(PRE_EXISTING_DATABASE_ID).read().block().getDatabase();
+        CosmosContainerSettings containerSettings = new CosmosContainerSettings(partitionedCollectionId, "/mypk");
+        CosmosContainer createdContainer = createdDatabase.createContainer(containerSettings).block().getContainer();
+        CosmosItem createdItem = createdContainer.createItem(new CosmosItem("{'id':'" + IdOfDocumentWithNoPk + "'}")).block().getItem();
+        Mono<CosmosItemResponse> readMono = createdItem.read(PartitionKey.None);
```
pradeeptamma: The value here is based on the container, so you should not directly use it here but go through the container.

moderakh:
This is exposing internal API as public API @srinathnarayanan can we talk when you get a chance?

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
```diff
@@ -134,6 +136,27 @@ public void setPaths(List<String> paths) {
...
+        return this.systemKey;
+    }
+
+    public PartitionKeyInternal getNonePartitionKeyValue() {
+        if (this.getPaths().size() == 0 || this.isSystemKey()) {
+            return PartitionKeyInternal.Empty;
+        } else {
+            return PartitionKeyInternal.UndefinedPartitionKey;
+        }
+    }
```
pradeeptamma: Move this to CosmosContainer.java

sdk/src/main/java/com/microsoft/azure/cosmos/CosmosClient.java
```diff
@@ -249,4 +253,8 @@ public CosmosDatabase getDatabase(String id) {
     public void close() {
         asyncDocumentClient.close();
     }
+
+    public Mono<DatabaseAccount> getDatabaseAccount() {
```
moderakh: why this should be public API?

sdk/src/main/java/com/microsoft/azure/cosmos/CosmosConfiguration.java
```diff
@@ -154,6 +154,42 @@ public CosmosConfiguration build() {
 
             return new CosmosConfiguration(this);
         }
+        
+        public URI getServiceEndpoint() {
+            return serviceEndpoint;
+        }
+
+        public String getKeyOrResourceToken() {
```
moderakh: why this should be public API?

sdk/src/main/java/com/microsoft/azure/cosmos/CosmosContainer.java
```diff
@@ -558,6 +575,15 @@ public CosmosItem getItem(String jsonString){
                                                                                                       response.getResponseHeaders()))));
     }
 
+    /**
+     * Gets a CosmosTrigger object without making a service call.
+     * @param jsonString body of the trigger in JSON.
+     * @return a cosmos trigger.
+     */
+    public CosmosTrigger getTrigger(String jsonString){
```
moderakh: I thought we gonna get rid of triggers for v3. No?

christopheranderson: No triggers are still around

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
srinathnarayanan: yes, will default to false if it is not returned from the backend

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
srinathnarayanan: CosmosContainer has no context of partitionKeyDefinitions (That would be in CosmosContainerSettings). Instead, I am going to make this package private and keep it inside the PartitionKeyDefinition.

commons/src/main/java/com/microsoft/azure/cosmosdb/PartitionKeyDefinition.java
srinathnarayanan: going to move this to CosmosBridgeInternal

christopheranderson:
@srinathnarayanan & @kushagraThapar will clean up any lingering test issues in another PR, mostly need this in for the OM changes.

PR #108, opened 2019-05-09
Refactoring V3 SDK Client

srinathnarayanan:
- Removed V3 sdk dependency on AsyncDocumentClient, RxDocumentClientImpl
- Replaced above dependencies with CosmosClientInternal and CosmosClientContext respectively
- Modified RxPartitionkeyRangeCache and ChangeFeedQueryImpl to no longer take any of the clients or client implementations. Instead takes the functions as parameters.
- Made RxGatewayStoreModel and ChangeFeedQueryImpl public so that they can be accessed by V3 SDK package

### DIFF
--- a/sdk/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ChangeFeedQueryImpl.java
+++ b/sdk/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ChangeFeedQueryImpl.java
@@ -43,22 +43,22 @@ import rx.Single;
 import rx.functions.Func1;
 import rx.functions.Func2;
 
-class ChangeFeedQueryImpl<T extends Resource> {
+public class ChangeFeedQueryImpl<T extends Resource> {
 
     private static final String IfNonMatchAllHeaderValue = "*";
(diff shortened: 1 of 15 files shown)


sdk/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ChangeFeedQueryImpl.java
```diff
@@ -43,22 +43,22 @@
 import rx.functions.Func1;
 import rx.functions.Func2;
 
-class ChangeFeedQueryImpl<T extends Resource> {
+public class ChangeFeedQueryImpl<T extends Resource> {
```
moderakh: >public [](start = 0, length = 6)

why did we make this public?

srinathnarayanan: wouldn't be accessible to com.microsoft.azure.cosmos package, where all the V3 stuff resides.

christopheranderson:
Confirmed with @srinathnarayanan he wants to abandon this for a different approach

PR #111, opened 2019-05-10
Retrieving query info using new query plan retriever

mbhaskar:
Removing proxy execution context
Retrieving query info using new query plan retriever

---
This change is [Reviewable](https://reviewable.io/reviews/azure/azure-cosmosdb-java/111)


gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/ProxyDocumentQueryExecutionContext.java
```diff
@@ -1,191 +0,0 @@
...
-/**
- * While this class is public, but it is not part of our published public APIs.
- * This is meant to be internally used only by our sdk.
- * 
- * This class is used as a proxy to wrap the
- * DefaultDocumentQueryExecutionContext which is needed for sending the query to
- * Gateway first and then uses PipelinedDocumentQueryExecutionContext after it
- * gets the necessary info.
- */
-public class ProxyDocumentQueryExecutionContext<T extends Resource> implements IDocumentQueryExecutionContext<T> {
```
moderakh: You can't imagine how happy I am that we have this new design and we are getting rid of the Proxy Query Execution Context. Thanks :-)

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
```diff
@@ -74,31 +75,52 @@
...
+        if (ResourceType.Document != resourceTypeEnum
+                || (feedOptions != null && feedOptions.getPartitionKey() != null)
+                || (feedOptions != null && feedOptions.getPartitionKeyRangeIdInternal() != null)) {
+            return Observable.just(queryExecutionContext);
+        }
+
+        Single<PartitionedQueryExecutionInfo> queryExecutionInfoSingle =
+                QueryPlanRetriever.getQueryPlanThroughGatewayAsync(client, query, resourceLink);
+
+        return collectionObs.flatMap(collection -> queryExecutionInfoSingle.toObservable()
```
moderakh: we are doing multiple Single to Observable conversion here in this block of code.

I wonder if instead we can do collectionObs.toSingle().flatMap(.) and inside the block we drop all the Single to Observable conversion?

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/BackPressureTest.java
```diff
@@ -101,7 +102,7 @@ public void readFeed() throws Exception {
             // validate that the difference between the number of requests to backend
             // and the number of returned results is always less than a fixed threshold
             assertThat(client.httpRequests.size() - subscriber.getOnNextEvents().size())
-                .isLessThanOrEqualTo(RxRingBuffer.SIZE);
+                .isLessThanOrEqualTo(RxRingBuffer.SIZE + THRESHOLD);
```
christopheranderson: This is just temporary until we move to Reactor right? If so, we should leave a TODO with a work item

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/ParallelDocumentQueryTest.java
```diff
@@ -233,12 +233,16 @@ public void crossPartitionQueryNotEnabled() {
...
+        List<Document> expectedDocs = createdDocuments;
 
-        FailureValidator validator = new FailureValidator.Builder()
-                .instanceOf(DocumentClientException.class)
-                .statusCode(400)
+        FeedResponseListValidator<Document> validator = new FeedResponseListValidator.Builder<Document>()
+                .totalSize(expectedDocs.size())
+                .exactlyContainsInAnyOrder(expectedDocs.stream().map(d -> d.getResourceId()).collect(Collectors.toList()))
+                .allPagesSatisfy(new FeedResponseValidator.Builder<Document>()
+                        .requestChargeGreaterThanOrEqualTo(1.0).build())
```
christopheranderson: Do we have a better guess than `1.0` for what the charge should be? We've had issues (in other SDKs mostly) with request charge being off for x-partition queries.

christopheranderson:
Also, please create an issue/PR to track the v3 port of this change before we merge.

moderakh:
Please see if you can remove the multiple addition single toObservable in the new block of code. If you could rely on Single.toFlatMap() that would be better and you can do less transformation.

Also you have test failure. other than that looks good.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
```diff
@@ -74,31 +75,52 @@
...
-                        resourceType,
-                        query,
-                        feedOptions,
-                        resourceLink,
-                        collection,
-                        isContinuationExpected,
-                        correlatedActivityId));
-
-        return proxyQueryExecutionContext;
+        DefaultDocumentQueryExecutionContext<T> queryExecutionContext =  new DefaultDocumentQueryExecutionContext<T>(
```
bchong95: Can we defer the allocation of this context until the check below says we need to?

In fact you can avoid the early exit entirely like this:

```
IDocumentQueryExecutionContext documentQueryExecutionContext;

if (is_resource_metadata_query){
    documentQueryExecutionContext = new DefaultDocumentQueryExecutionContext()
}
else{
    documentQueryExecutionContext = CreateSpecializedDocumentQueryExecutionContext();
}
```

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
```diff
@@ -74,31 +75,52 @@
...
+                resourceTypeEnum,
+                resourceType,
+                query,
+                feedOptions,
+                resourceLink,
+                correlatedActivityId,
+                isContinuationExpected);
+
+        if (ResourceType.Document != resourceTypeEnum
+                || (feedOptions != null && feedOptions.getPartitionKey() != null)
```
bchong95: We don't want queries going to default even if the user supplies a partition key, since queries like GROUP BY and DISTINCT won't work even if it's on a single partition.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
```diff
@@ -114,6 +136,10 @@
             String collectionRid,
             UUID correlatedActivityId) {
 
+        if(feedOptions == null){
```
bchong95: nit keep the spacing consistent.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
```diff
@@ -114,6 +136,10 @@
...
+            feedOptions = new FeedOptions();
```
bchong95: Why do we need to replace with a new feed options? Is `null` not a valid one?

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
```diff
@@ -124,6 +150,23 @@
 
         QueryInfo queryInfo = partitionedQueryExecutionInfo.getQueryInfo();
 
+        // non value aggregates must go through DefaultDocumentQueryExecutionContext
```
bchong95: Move this code outside of here. It makes more sense for it to be in the factory, since at this point we should just commit to making a pipelined execution context.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
```diff
@@ -124,6 +150,23 @@
...
+        if(queryInfo.hasAggregates() && !queryInfo.hasSelectValue()){
```
bchong95: Please leave a comment explaining that this is to avoid a breaking change.

A single partition query can serve queries like:

`SELECT AVG(c.age) FROM c`
`SELECT MIN(c.age) + 5 FROM c`
`SELECT MIN(c.age), MAX(c.age) FROM c` 

while pipelined queries can only serve

`SELECT VALUE <AGGREGATE>` so we send the query down the old pipeline to avoid a breaking change. In the v3 SDK we will skip this check.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/QueryFeature.java
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.azure.cosmosdb.rx.internal.query;
+
+enum QueryFeature {
```
bchong95: Is it possible to make this a flags enums? If it's a lot of work we can skip it since it's not that important

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/QueryPlanRetriever.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+
+class QueryPlanRetriever {
+    private static final String TRUE = "True";
+    private static final String QUERY_VERSION = "1.0";
+    private static final String supportedQueryFeatures = QueryFeature.Aggregate.name() + ", " +
+            QueryFeature.CompositeAggregate.name() + "," +
```
bchong95: We don't want to support composite aggregate. Please add a test for the unsupported features (Composite Aggregate, Multiple Aggregate, GROUP BY). I added one to the v3 SDK if you want to just port that.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/BackPressureTest.java
bchong95: I think we might we able to lower the threshold to just 1, since it's just one additional request for the plan.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
bchong95: Please add a test to stress this. Set Max Item Count to 1 and have duplicate values and see if DISTINCT still works.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/BackPressureTest.java
mbhaskar: This is due to backpressure and behaving a little differently and is off by 2. Added 5 just to be safe. We will revisit this as part of Reactor conversion #125

mbhaskar:
> Also, please create an issue/PR to track the v3 port of this change before we merge.

#126

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
mbhaskar: Moved

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
mbhaskar: Done

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/QueryFeature.java
mbhaskar: We can by adding a constructor. Looked a bit messy so left it like this

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/QueryPlanRetriever.java
mbhaskar: Added tests

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
mbhaskar: We are using same documentQueryExecutionContext for bypassing no value aggregates and also to getTargetPartitionKeyRanges. So creating it early. And doing early exit as I already have the context.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
mbhaskar: Support for this is added in RxJava2 (https://github.com/ReactiveX/RxJava/issues/4635) but cant flatmap single to observable in rx1.

moderakh: ok thanks. Then please mark this as resolved.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
mbhaskar: As we are trying to use feedoptions below to get intialpagesize. We used to do the same before in proxy

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java
mbhaskar: will add tests in pr for distinct

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/ParallelDocumentQueryTest.java
mbhaskar: Most of the tests use 1.0. Will try a better value in V3

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/ParallelDocumentQueryTest.java
```diff
@@ -435,4 +440,31 @@ private void queryWithContinuationTokensAndPageSizes(String query, int[] pageSiz
 
         return receivedDocuments;
     }
+
+    @Test(groups = { "simple" }, timeOut = TIMEOUT)
+    public void unsupportedQueries() {
```
bchong95: These queries should be supported for non parallel queries. Can we test for this?

mbhaskar: Will add in subsequent query pr for skipntake

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/routing/RoutingMapProviderHelper.java
```diff
@@ -94,4 +98,58 @@ private static String max(String left, String right) {
 
         return targetRanges;
     }
+
+    public static Single<List<PartitionKeyRange>> getOverlappingRanges(IRoutingMapProvider routingMapProvider,
+                                                                       String resourceId, List<Range<String>> sortedRanges) {
+
+        if (sortedRanges == null) {
```
bchong95: null check on `routingMapProvider`?

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/routing/RoutingMapProviderHelper.java
```diff
@@ -94,4 +98,58 @@ private static String max(String left, String right) {
...
+        List<PartitionKeyRange> targetRanges = new ArrayList<>();
+        final Iterator<Range<String>> iterator = sortedRanges.iterator();
+
+        return Observable.defer(() -> {
+            if (!iterator.hasNext()) {
+                return Observable.empty();
+            }
+
+            Range<String> queryRange;
+            Range<String> tSortedRange = iterator.next();
```
bchong95: `tSortedRange` is that a typo?

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/routing/RoutingMapProviderHelper.java
```diff
@@ -94,4 +98,58 @@ private static String max(String left, String right) {
 
         return targetRanges;
     }
+
+    public static Single<List<PartitionKeyRange>> getOverlappingRanges(IRoutingMapProvider routingMapProvider,
```
moderakh: we need some unit testing for this (not end to end testing).

Please add some tests mocking the `IRoutingMapProvider` behaviour and validate the expected result by `getOverlappingRanges()`

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/routing/RoutingMapProviderHelper.java
mbhaskar: corrected

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/QueryPlanRetriever.java
```diff
@@ -0,0 +1,71 @@
...
+import com.microsoft.azure.cosmosdb.rx.internal.RxDocumentServiceRequest;
+import rx.Single;
+
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+
+class QueryPlanRetriever {
+    private static final String TRUE = "True";
+    private static final String QUERY_VERSION = "1.0";
```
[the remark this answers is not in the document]
mbhaskar: Moved

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/ParallelDocumentQueryExecutionContextBase.java
```diff
@@ -84,11 +84,18 @@ protected void initialize(String collectionRid,
         Map<String, String> commonRequestHeaders = createCommonHeadersAsync(this.getFeedOptions(null, null));
 
         for (PartitionKeyRange targetRange : partitionKeyRangeToContinuationTokenMap.keySet()) {
+
             Func3<PartitionKeyRange, String, Integer, RxDocumentServiceRequest> createRequestFunc = (partitionKeyRange,
                     continuationToken, pageSize) -> {
                 Map<String, String> headers = new HashMap<>(commonRequestHeaders);
                 headers.put(HttpConstants.HttpHeaders.CONTINUATION, continuationToken);
                 headers.put(HttpConstants.HttpHeaders.PAGE_SIZE, Strings.toString(pageSize));
+                if(feedOptions.getPartitionKey() != null){
```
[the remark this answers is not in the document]
mbhaskar: partition key takes higher priority than partition key range id. We are doing this in .NEt too

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/QueryPlanRetriever.java
```diff
@@ -0,0 +1,71 @@
...
+    static Single<PartitionedQueryExecutionInfo> getQueryPlanThroughGatewayAsync(IDocumentQueryClient queryClient,
+                                                                                 SqlQuerySpec sqlQuerySpec, String resourceLink) {
+        Map<String, String> requestHeaders = new HashMap<>();
+        requestHeaders.put(HttpConstants.HttpHeaders.CONTENT_TYPE, RuntimeConstants.MediaTypes.JSON);
+        requestHeaders.put(HttpConstants.HttpHeaders.IS_QUERY_PLAN_REQUEST, TRUE);
+        requestHeaders.put(HttpConstants.HttpHeaders.SUPPORTED_QUERY_FEATURES, SUPPORTED_QUERY_FEATURES);
+        requestHeaders.put(HttpConstants.HttpHeaders.QUERY_VERSION, QUERY_VERSION);
+
+        RxDocumentServiceRequest request;
+        request = RxDocumentServiceRequest.create(OperationType.QueryPlan,
```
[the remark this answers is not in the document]
mbhaskar: Done

commons/src/main/java/com/microsoft/azure/cosmosdb/FeedOptions.java
```diff
@@ -229,4 +231,19 @@ public void setResponseContinuationTokenLimitInKb(int limitInKb) {
...
+     */
+    public boolean getAllowEmptyPages() {
+        return allowEmptyPages;
+    }
+
+    /**
+     * Sets the option to allow empty result pages in feed response.
+     * @param allowEmptyPages
+     */
+    public void setAllowEmptyPages(boolean allowEmptyPages) {
```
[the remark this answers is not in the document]
mbhaskar: false. Added to javadoc

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/routing/RoutingMapProviderHelper.java
mbhaskar: Added unit test

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/routing/RoutingMapProviderHelper.java
mbhaskar: Done

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/AggregateQueryTests.java
```diff
@@ -94,8 +94,14 @@ public void queryDocumentsWithAggregates(boolean qmEnabled) throws Exception {
...
-
-        for (QueryConfig queryConfig : queryConfigs) {    
+        
+        for (QueryConfig queryConfig : queryConfigs) {
+            // Cross partition Non value aggregates are not supported
+            if(queryConfig.query.contains("VALUE")){
+                options.setEnableCrossPartitionQuery(true);
+            }else{
+                options.setEnableCrossPartitionQuery(false);
+            }
```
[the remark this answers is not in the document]
mbhaskar: queryConfigs has queries with and without Value so both paths are covered

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/QueryPlanRetriever.java
```diff
@@ -0,0 +1,71 @@
...
+                resourceLink,
+                requestHeaders);
+        request.UseGatewayMode = true;
+        request.setContentBytes(sqlQuerySpec.toJson().getBytes(StandardCharsets.UTF_8));
+
+        return queryClient.executeQueryAsync(request).flatMap(rxDocumentServiceResponse -> {
+            PartitionedQueryExecutionInfo partitionedQueryExecutionInfo =
+                    new PartitionedQueryExecutionInfo(rxDocumentServiceResponse.getReponseBodyAsString());
+            return Single.just(partitionedQueryExecutionInfo);
+        });
```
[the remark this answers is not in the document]
mbhaskar: This call is made internally in query so the query retry should handle this too

PR #128, opened 2019-05-17
Implementation for Change Feed Processor

milismsft:
Refactoring from V2 Rx based code to Reactor Flux/Mono.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/azure-cosmosdb-java/128)


moderakh:
1. implementation doesn't use async pattern, by using `block()` or `ExecutorService` you are not using the async nature of the async sdk
2. whoever subscribes to `Flux` and `Mono` should be able to unsubscribe using reactor stream unsubscribe/cancel https://www.reactive-streams.org/reactive-streams-1.0.2-javadoc/org/reactivestreams/Subscription.html#cancel-- why aren't we using this instead of CancellationToken defined in the code?

christopheranderson:
https://github.com/Azure/azure-cosmosdb-java/issues/173 tracks some fixes we need to do before GA

christopheranderson:
Created issue to track cleaning up implementation: https://github.com/Azure/azure-cosmosdb-java/issues/175

PR #136, opened 2019-05-22
Fix for global strong retry

moderakh:
* Fix for global strong retry
* improved test setup time by doing bulk insert in parallel instead of serial


christopheranderson:
Closed in favor of #196

PR #192, opened 2019-06-17
Adding conflict api

mbhaskar:
Fixing Permission tests
Making id method of Resource package private and adding id on
individual resource classes to enable fluent composition

### DIFF
--- a/commons/src/main/java/com/azure/data/cosmos/PartitionKeyRange.java
+++ b/commons/src/main/java/com/azure/data/cosmos/PartitionKeyRange.java
@@ -54,6 +54,16 @@ public class PartitionKeyRange extends Resource {
         super(jsonString);
     }
 
+    /**
+     * Set id of partition key range
+     * @param id the name of the resource.
+     * @return the partition key range
+     */
+    public PartitionKeyRange id(String id) {
+        super.id(id);
+        return this;
+    }
+
     public PartitionKeyRange(String id, String minInclusive, String maxExclusive) {
         super();
         this.id(id);
--- a/sdk/src/test/java/com/azure/data/cosmos/rx/CosmosConflictTest.java
+++ b/sdk/src/test/java/com/azure/data/cosmos/rx/CosmosConflictTest.java
@@ -1,0 +1,98 @@
+/*
+ * The MIT License (MIT)
+ * Copyright (c) 2018 Microsoft Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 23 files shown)


sdk/src/test/java/com/azure/data/cosmos/rx/PermissionCrudTest.java
```diff
@@ -207,21 +221,17 @@ public void replacePermission() throws Exception {
     @BeforeClass(groups = { "simple" }, timeOut = SETUP_TIMEOUT)
     public void beforeClass() {
         client = clientBuilder().build();
-        createdDatabase = SHARED_DATABASE;
+        createdDatabase = createDatabase(client, databaseId);
```
kushagraThapar: Why are we not using SHARED_DATABASE ?

mbhaskar: I used the same pattern from Container and User tests. @srinathnarayanan any particular reason?

sdk/src/test/java/com/azure/data/cosmos/rx/CosmosConflictTest.java
```diff
@@ -0,0 +1,98 @@
...
+            numberOfResults += pageSize;
+        }
+        assertThat(numberOfResults).isEqualTo(expectedNumberOfConflicts);
+    }
+
+
+    @BeforeClass(groups = { "simple" }, timeOut = SETUP_TIMEOUT)
+    public void beforeClass() {
+        client = clientBuilder().build();
+        createdCollection = getSharedMultiPartitionCosmosContainer(client);
```
moderakh: that's great that we are using the shared collection.

sdk/src/test/java/com/azure/data/cosmos/rx/PermissionCrudTest.java
moderakh: in general we should try to use shared database and shared collection, however I think for permissions it might be a bit tricky, because permissions are created in the database level, and after the tests finish I am afraid we may not have a way of cleaning all permissions. 

So for Users and Permissions tests it is ok to create a new database.

commons/src/main/java/com/azure/data/cosmos/PartitionKeyRange.java
```diff
@@ -54,6 +54,16 @@ public PartitionKeyRange(String jsonString) {
         super(jsonString);
     }
 
+    /**
+     * Set id of partition key range
+     * @param id the name of the resource.
+     * @return the partition key range
+     */
+    public PartitionKeyRange id(String id) {
```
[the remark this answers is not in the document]
mbhaskar: The class and all the methods are still public. Hiding this would be a different PR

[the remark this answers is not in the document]
moderakh: ok sounds good. Please just create a work item for that so we remember to do before GA.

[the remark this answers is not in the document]
mbhaskar: #193

PR #204, opened 2019-06-25
Hide Json Serializable

kushagraThapar:
Reduced scope of set and remove method of JsonSerializable from public to package-private.

Created BridgeInternal methods for these. 

Reduced scope of methods returning objectMapper, objectNode, jsonNode, etc.

### DIFF
--- a/benchmark/src/main/java/com/azure/data/cosmos/benchmark/AsyncBenchmark.java
+++ b/benchmark/src/main/java/com/azure/data/cosmos/benchmark/AsyncBenchmark.java
@@ -24,6 +24,7 @@
 package com.azure.data.cosmos.benchmark;
 
 import com.azure.data.cosmos.AsyncDocumentClient;
+import com.azure.data.cosmos.BridgeInternal;
 import com.azure.data.cosmos.Database;
 import com.azure.data.cosmos.Document;
 import com.azure.data.cosmos.DocumentCollection;
@@ -100,12 +101,12 @@ abstract class AsyncBenchmark<T> {
                 String uuid = UUID.randomUUID().toString();
                 Document newDoc = new Document();
                 newDoc.id(uuid);
-                newDoc.set(partitionKey, uuid);
-                newDoc.set("dataField1", dataFieldValue);
-                newDoc.set("dataField2", dataFieldValue);
-                newDoc.set("dataField3", dataFieldValue);
-                newDoc.set("dataField4", dataFieldValue);
-                newDoc.set("dataField5", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, partitionKey, uuid);
+                BridgeInternal.setProperty(newDoc, "dataField1", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField2", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField3", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField4", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField5", dataFieldValue);
                 Flux<Document> obs = client.createDocument(collection.selfLink(), newDoc, null, false)
                                                  .map(ResourceResponse::getResource);
                 createDocumentObservables.add(obs);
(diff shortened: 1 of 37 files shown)


benchmark/src/main/java/com/azure/data/cosmos/benchmark/AsyncBenchmark.java
```diff
@@ -100,12 +101,12 @@
...
-                newDoc.set("dataField2", dataFieldValue);
-                newDoc.set("dataField3", dataFieldValue);
-                newDoc.set("dataField4", dataFieldValue);
-                newDoc.set("dataField5", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, partitionKey, uuid);
+                BridgeInternal.setProperty(newDoc, "dataField1", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField2", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField3", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField4", dataFieldValue);
+                BridgeInternal.setProperty(newDoc, "dataField5", dataFieldValue);
```
moderakh: hmm, there has to proper way for setting entries in the document.

otherwise the customer may use this as an example which kind of defeats the purpose.

benchmark/src/main/java/com/azure/data/cosmos/benchmark/AsyncMixedBenchmark.java
```diff
@@ -57,12 +58,12 @@ protected void performWorkload(BaseSubscriber<Document> documentBaseSubscriber,
...
-            newDoc.set("dataField2", dataFieldValue);
-            newDoc.set("dataField3", dataFieldValue);
-            newDoc.set("dataField4", dataFieldValue);
-            newDoc.set("dataField5", dataFieldValue);
+            BridgeInternal.setProperty(newDoc, partitionKey, idString);
+            BridgeInternal.setProperty(newDoc, "dataField1", dataFieldValue);
+            BridgeInternal.setProperty(newDoc, "dataField2", dataFieldValue);
+            BridgeInternal.setProperty(newDoc, "dataField3", dataFieldValue);
+            BridgeInternal.setProperty(newDoc, "dataField4", dataFieldValue);
+            BridgeInternal.setProperty(newDoc, "dataField5", dataFieldValue);
```
moderakh: ditto and everywhere else in the benchamrk tool

benchmark/src/main/java/com/azure/data/cosmos/benchmark/AsyncBenchmark.java
moderakh: a lot of customers use benchmark tool as a sample code. this is not really internally exposed code. think of it as sample code also.

examples/src/main/java/com/azure/data/cosmos/rx/examples/multimaster/samples/ConflictWorker.java
```diff
@@ -498,8 +499,8 @@ public void runDeleteConflictOnUdp() throws Exception {
     private Flux<Document> tryInsertDocument(AsyncDocumentClient client, String collectionUri, Document document, int index) {
 
         logger.debug("region: {}", client.getWriteEndpoint());
-        document.set("regionId", index);
-        document.set("regionEndpoint", client.getReadEndpoint());
+        BridgeInternal.setProperty(document, "regionId", index);
+        BridgeInternal.setProperty(document, "regionEndpoint", client.getReadEndpoint());
```
moderakh: hmm, if we do this in sample code, then the customer will use this as sample.

benchmark/src/main/java/com/azure/data/cosmos/benchmark/AsyncBenchmark.java
kushagraThapar: The reason we want to get rid of JsonSerializable form the old surface as well, is because customer can call set methods in Document, and can set anything, whereas using BridgeInternal, at least we specifically say that don't use this class, as this is just for internal use. 

@christopheranderson  , what do you think about this change ?

moderakh: discussed offline, these are side effect.

PR #205, opened 2019-06-25
Cosmos Resource refactoring

christopheranderson:
- Removes Cosmos Resource
- Fixes some names and Java Docs
- Moves sprocs, udfs, and triggers

For review, it might be easier to look at each commit

---
This change is [Reviewable](https://reviewable.io/reviews/azure/azure-cosmosdb-java/205)


sdk/src/main/java/com/azure/data/cosmos/CosmosScripts.java
```diff
@@ -0,0 +1,293 @@
+package com.azure.data.cosmos;
```
moderakh: file header is missing. please add.

christopheranderson: done

PR #104, opened 2019-05-07
Updated parent pom to include collected artifacts

kushagraThapar:
Updated parent to include collected artifacts
Updated child sub-module pom to skip collection of artifacts when running maven antrun plugin

PR #206, opened 2019-06-25
2.5.0

christopheranderson:
2.5.0 release


christopheranderson:
FYI - this bringing in changes from #104 that we decided to kill. I took the minimum set of those changes to get the release packaging fixed up. It's non-ideal but it's functional.

PR #209, opened 2019-06-26
Moved old types to internal package and other changes

kushagraThapar:
Collection to list changes, moved old types to internal, pom file change to exclude BridgeInternal and CosmosBridgeInternal


commons/src/main/java/com/azure/data/cosmos/ConnectionPolicy.java
```diff
@@ -23,6 +23,8 @@
 
 package com.azure.data.cosmos;
 
+import com.azure.data.cosmos.internal.MediaReadMode;
```
christopheranderson: Seems like we accidentally added this here.

commons/src/main/java/com/azure/data/cosmos/JsonSerializable.java
```diff
@@ -381,7 +381,11 @@ public Double getDouble(String propertyName) {
                 } else if (isJsonSerializable) {
                     // JsonSerializable
                     try {
-                        result.add(c.getConstructor(String.class).newInstance(toJson(n)));
+                        Constructor<T> constructor = c.getDeclaredConstructor(String.class);
+                        if(Modifier.isPrivate(constructor.getModifiers())) {
+                            constructor.setAccessible(true);
```
moderakh: if they are our types we should make the constructors non private. that way you won't need to override access modifiers.

commons/src/main/java/com/azure/data/cosmos/JsonSerializable.java
```diff
@@ -103,7 +104,7 @@ private static void checkForValidPOJO(Class<?> c) {
         }
     }
 
-    Logger getLogger() {
+    public Logger getLogger() {
```
moderakh: why public?

commons/src/main/java/com/azure/data/cosmos/JsonSerializable.java
```diff
@@ -309,7 +310,11 @@ public Double getDouble(String propertyName) {
                 }
             } else if (JsonSerializable.class.isAssignableFrom(c)) {
                 try {
-                    return c.getConstructor(String.class).newInstance(toJson(jsonObj));
+                    Constructor<T> constructor = c.getDeclaredConstructor(String.class);
+                    if(Modifier.isPrivate(constructor.getModifiers())) {
+                        constructor.setAccessible(true);
```
moderakh: if they are our types we should make the constructors non private. that way you won't need to override access modifiers.

commons/src/main/java/com/azure/data/cosmos/SpatialIndex.java
```diff
@@ -67,7 +67,7 @@ public SpatialIndex(String jsonString) {
      *
      * @return the data type.
      */
-    public DataType dataType() {
+    private DataType dataType() {
```
moderakh: don't we need these getter and setters?

commons/src/test/java/com/azure/data/cosmos/DocumentCollectionTests.java
```diff
@@ -52,4 +55,22 @@ public void getPartitionKey_serializeAndDeserialize()  {
         assertThat(parsedColl.getPartitionKey().paths()).isEqualTo(partitionKeyDefinition.paths());
         assertThat(parsedColl.getPartitionKey().version()).isEqualTo(partitionKeyDefinition.version());
     }
+
+    @Test(groups = { "unit"})
+    public void IndexingPolicy_serializeAndDeserialize() {
```
moderakh: please change to lower camel case:

`IndexingPolicy_serializeAndDeserialize` -> `indexingPolicy_serializeAndDeserialize`

pom.xml
```diff
@@ -283,6 +283,8 @@
           <additionalOptions>${javadoc.opts}</additionalOptions>
           <sourceFileExcludes>
             <sourceFileExclude>**/internal/**/*.java</sourceFileExclude>
+            <sourceFileExclude>**/com.azure.data.cosmos.BridgeInternal.java</sourceFileExclude>
+            <sourceFileExclude>**/com.azure.data.cosmos.CosmosBridgeInternal.java</sourceFileExclude>
```
moderakh: I think here package name `com.azure.data.cosmos` should be translated to directory structure `com/azure/data/cosmos`.

The easier way is to just have this instead:
**/BridgeInternal.java

sdk/src/main/java/com/azure/data/cosmos/AsyncDocumentClient.java
```diff
@@ -624,7 +634,7 @@ public void setTokenResolver(TokenResolver tokenResolver) {
      * @return an {@link Flux} containing the single resource response with the created stored procedure or an error.
      */
     Flux<ResourceResponse<StoredProcedure>> createStoredProcedure(String collectionLink, StoredProcedure storedProcedure,
-                                                                        RequestOptions options);
+                                                                  RequestOptions options);
```
moderakh: remove empty line please

sdk/src/main/java/com/azure/data/cosmos/AsyncDocumentClient.java
```diff
@@ -871,7 +881,7 @@ public void setTokenResolver(TokenResolver tokenResolver) {
      * @return an {@link Flux} containing the single resource response with the created user defined function or an error.
      */
     Flux<ResourceResponse<UserDefinedFunction>> createUserDefinedFunction(String collectionLink, UserDefinedFunction udf,
-                                                                                RequestOptions options);
+                                                                          RequestOptions options);
```
moderakh: remove empty line please

commons/src/main/java/com/azure/data/cosmos/JsonSerializable.java
moderakh: if this class and the source classes are in the same package and if constructor of the class is package private this probably should work without the need for overriding accessiblity

pom.xml
kushagraThapar: Cool, will try that. Thanks.

commons/src/main/java/com/azure/data/cosmos/ConnectionPolicy.java
kushagraThapar: This has always been there, once I moved MediaReadMode to internal package, so it added the import.

commons/src/main/java/com/azure/data/cosmos/JsonSerializable.java
kushagraThapar: Agreed, but not touching any JsonSerializable logic given the time constraint. In the long run we plan to remove JsonSerializable completely and then we can re-write this logic.

commons/src/main/java/com/azure/data/cosmos/JsonSerializable.java
kushagraThapar: This logger was public even before, but made it package private in my last PR.
In this PR, since a lot of old types have been moved to internal package, which inherits this, so needed to make this public again to be accessible across packages.

commons/src/test/java/com/azure/data/cosmos/DocumentCollectionTests.java
kushagraThapar: Done

commons/src/main/java/com/azure/data/cosmos/SpatialIndex.java
kushagraThapar: They are not being used as of now, so made them private, but you are right, we might need them in the future, so will make them public again.

PR #210, opened 2019-06-26
Fixing executeStoredProcedureInternal to use client retry policy for …

mbhaskar:
…retries


gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -108,7 +108,7 @@ public ClientRetryPolicy(GlobalEndpointManager globalEndpointManager,
         if (clientException != null &&
                 Exceptions.isStatusCode(clientException, HttpConstants.StatusCodes.FORBIDDEN) &&
                 Exceptions.isSubStatusCode(clientException, HttpConstants.SubStatusCodes.DATABASE_ACCOUNT_NOTFOUND) &&
-                this.isReadRequest)
+                (this.isReadRequest || this.canUseMultipleWriteLocations))
```
moderakh: doesn't this affect non sproc operations?

mbhaskar: Yes this is not specific to sprocs. This check is to enable multimaster accounts to retry on 403 and 1008. .NET does the same

PR #222, opened 2019-07-04
Initialize options variable inside createContainerIfNotExists()

### LINKED ISSUE #221: Null pointer when using createContainerIfNotExists()
**Describe the bug**
I have a null pointer exception when doing `database.createContainerIfNotExists(new CosmosContainerProperties(CONTAINER_NAME, "/partition"))`

In fact this method calls this line:
https://github.com/Azure/azure-cosmosdb-java/blob/v3/sdk/src/main/java/com/azure/data/cosmos/CosmosDatabase.java#L250

Which uses a `null` value for the class `CosmosContainerRequestOptions`.

So when this line is called for creating a container: https://github.com/Azure/azure-cosmosdb-java/blob/v3/sdk/src/main/java/com/azure/data/cosmos/CosmosDatabase.java#L177 it does a `options.toRequestOptions()` which fails as the object is null.

**Environment summary**
SDK Version: 3.0.0
Java JDK version: openjdk version "11.0.3" 2019-04-16
OS Version (e.g. Windows, Linux, MacOSX): MacOSX

jdubois:
Fix #221

### DIFF
--- a/sdk/src/main/java/com/azure/data/cosmos/CosmosDatabase.java
+++ b/sdk/src/main/java/com/azure/data/cosmos/CosmosDatabase.java
@@ -246,8 +246,9 @@ public class CosmosDatabase {
      *         created or existing container or an error.
      */
     public Mono<CosmosContainerResponse> createContainerIfNotExists(CosmosContainerProperties containerSettings) {
+        CosmosContainerRequestOptions options = new CosmosContainerRequestOptions();
         CosmosContainer container = getContainer(containerSettings.id());
-        return createContainerIfNotExistsInternal(containerSettings, container, null);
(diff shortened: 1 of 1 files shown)


christopheranderson:
I'm fine with this as is, but ideally we'd fix it such that options isn't required to be initialized. We already create a lot of objects per request.

christopheranderson:
(btw, we have some weird timeout bugs in our tests, so don't worry about the failures. We'll get that cleaned up next week.)

moderakh:
development of v3 sdk is moved to https://github.com/Azure/azure-sdk-for-java
https://github.com/Azure/azure-sdk-for-java/pull/4488

This repo will only be used for v2 sdk development. closing it here.

PR #224, opened 2019-07-08
Fixed encoding to ensure lower case % encoding

srinathnarayanan:
URL encoded auth key will not contain charatcers like "%2B" but instead "%2b". This is needed since the backend does case sensitive decoding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ hs_err_pid*
 .classpath
 .idea
 *.iml
+*/test-output/
 
 doc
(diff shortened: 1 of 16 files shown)


srinathnarayanan:
> Could you quickly add a test case that has the offending string that we struggled with before?

The offending string is the Auth header. I have added tests that perform a bunch of operations and check that the auth header being sent over the wire has the required encoding.

christopheranderson:
No longer necessary

PR #230, opened 2019-07-16
memory improvements

moderakh:
This PR Includes the following memory improvements. To review it is easier to review the commit per improvement:

(The following numbers are based on running benchmark tool on 16cpu core Linux VM with ReadThroughput workload and 4000 as degree of concurrency.)

1) when reading from network wire (through netty ByteBuf) we used to copy data once to ByteArrayOutputStream and from that we created String. Later ByteArrayOutputStream will get garbage collected, on high load this can lead to GCing 20-30% unnecessary data. The improvement is to pool the buffers: https://github.com/Azure/azure-cosmosdb-java/pull/230/commits/077d2ce24001726a5cb4399fbc8dbc187cdcec7e

2) we were deserializing replica physicalAddress in the AddressSelector to URI per request, we can cache the parsed URI form of the replica physical Address, also in the HttpTransportClient we were using URI for constructing paths which consumes more memory comparing to pure String. This 
somewhat contributed 5-10% memory which gets GCed:
https://github.com/Azure/azure-cosmosdb-java/pull/230/commits/bd888d506998f93757eeef19a74da13350e537b6

3) we were decoding/encoding UUID to String unnecessarily. 2% memory improvement:
https://github.com/Azure/azure-cosmosdb-java/pull/230/commits/d62a1aa208814d0e787fd37a74e2b682c954f4d7

4) in auth token generation we are using StringBuilder which grows (it copies data on each grow) as we append to it. This is improved by preserving memory 1-2% memory improvement:
https://github.com/Azure/azure-cosmosdb-java/pull/230/commits/c8fc483569bc2ea930ef462d92ded43044b95e23

### DIFF
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
@@ -1,0 +1,135 @@
+/*
+ * The MIT License (MIT)
+ * Copyright (c) 2018 Microsoft Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package com.microsoft.azure.cosmosdb.internal;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.concurrent.ConcurrentLinkedDeque;
+
+/**
+ * Pool of Byte Buffers, this helps in re-using memory
+ */
+public class ByteBufferPool {
+
+    public class ByteBufferWrapper {
+        private final PoolSegment poolSegment;
+        private final ByteBuffer byteBuffer;
+        private ByteBufferWrapper(ByteBuffer byteBuffer, PoolSegment poolSegment) {
+            this.byteBuffer = byteBuffer;
+            this.poolSegment = poolSegment;
+        }
+
+        public ByteBuffer getByteBuffer() {
+            return this.byteBuffer;
+        }
+    }
+
+    private class PoolSegment {
+        public PoolSegment(int byteBufferCount, ConcurrentLinkedDeque<ByteBufferWrapper> byteBuffersPool) {
+            this.byteBufferSize = byteBufferCount;
+            this.byteBuffersPool = byteBuffersPool;
+        }
+
+        private final int byteBufferSize;
+        private final ConcurrentLinkedDeque<ByteBufferWrapper> byteBuffersPool;
+    }
+
+    private final ArrayList<PoolSegment> poolSegmentList = new ArrayList<>();
+
+
+    private final static ByteBufferPool instant = new ByteBufferPool();
+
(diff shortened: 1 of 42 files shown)


commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
```diff
@@ -0,0 +1,135 @@
...
+            this.poolSegment = poolSegment;
+        }
+
+        public ByteBuffer getByteBuffer() {
+            return this.byteBuffer;
+        }
+    }
+
+    private class PoolSegment {
+        public PoolSegment(int byteBufferCount, ConcurrentLinkedDeque<ByteBufferWrapper> byteBuffersPool) {
```
kushagraThapar: We should just name this same as the instance variable, byteBufferSize.

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
```diff
@@ -0,0 +1,135 @@
...
+    private final static ByteBufferPool instant = new ByteBufferPool();
+
+    public static ByteBufferPool getInstant() {
+        return instant;
+    }
+
+    private final Logger logger = LoggerFactory.getLogger(ByteBufferPool.class);
+
+    private ByteBufferPool() {
+        logger.info("Initializing ByteBuffer Pool");
```
kushagraThapar: Do we want these logs info instead of debug or trace ?

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
```diff
@@ -0,0 +1,135 @@
...
+        long totalSize = 0;
+
+        for(int byteBufferSize = 1024, segmentSize = 1024; segmentSize > 0; byteBufferSize *= 2, segmentSize /= 2) {
+            logger.debug("Creating pool segment: ByteBuffer size {}, pool segment size {}", byteBufferSize, segmentSize);
+            poolSegmentList.add(createByteBufferPoolSegment(byteBufferSize, segmentSize));
+            totalSize += (byteBufferSize * segmentSize);
+        }
+
+        logger.info("Total ByteBuffer Pool Size {}", totalSize);
```
kushagraThapar: Same here.

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/HttpUtils.java
```diff
@@ -65,13 +65,8 @@ public static String urlDecode(String url) {
...
     }
 
-    public static URI toURI(String uri) {
-        try {
-            return new URI(uri);
-        } catch (Exception e) {
-            log.error("failed to parse {}", uri, e);
-            throw new IllegalArgumentException("failed to parse uri " + uri, e);
-        }
+    public static Uri toURI(Uri uri) {
```
kushagraThapar: We should rename this to toUri since we have moved away from URI.

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/ResponseUtils.java
```diff
@@ -23,43 +23,49 @@
...
+                                out.getByteBuffer().limit(limit);
+                                bb.readBytes(out.getByteBuffer());
+                                assert contentLength == limit :
+                                        "contentLength " + contentLength + " is not equal to ByteBuffer readable bytes" + limit;
+
                                 return out;
-                            } catch (IOException e) {
-                                throw new RuntimeException(e);
+                            } catch (Throwable t) {
+                                ByteBufferPool.getInstant().release(byteBufferWrapper);
```
kushagraThapar: A good practice would be to do this in finally block as you have done in the below code.

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
moderakh: thanks. will be in the next commit.

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/HttpUtils.java
moderakh: sure. thank you. will rename

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/ResponseUtils.java
moderakh: you are right. this is actually a bug in the code, which later got fixed. see this commit:
https://github.com/Azure/azure-cosmosdb-java/pull/230/commits/44b47c53e55cba37ee6a7ba8471996691c0a0b06

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/HttpUtils.java
moderakh: that method is removed as it is no-op now.

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
moderakh: thanks. changed to debug.

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ByteBufferPool.java
moderakh: changed to debug.

PR #234, opened 2019-07-22
Offset limit query support

mbhaskar:
Adding support to offset limit queries 
This also supports offset-limit queries with continuation token

### DIFF
--- a/gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/query/QueryInfo.java
+++ b/gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/query/QueryInfo.java
@@ -41,6 +41,8 @@ public final class QueryInfo extends JsonSerializable {
     private Collection<AggregateOperator> aggregates;
     private Collection<String> orderByExpressions;
     private String rewrittenQuery;
+    private Integer offset;
+    private Integer limit;
 
     public QueryInfo() { }
 
@@ -90,8 +92,24 @@ public final class QueryInfo extends JsonSerializable {
                 ? this.orderByExpressions
                 : (this.orderByExpressions = super.getCollection("orderByExpressions", String.class));
     }
-    
-    public boolean hasSelectValue(){
+
+    public boolean hasSelectValue() {
         return super.has(HAS_SELECT_VALUE) && super.getBoolean(HAS_SELECT_VALUE);
+    }
+
+    public boolean hasOffset() {
+        return this.getOffset() != null;
+    }
+
+    public boolean hasLimit() {
+        return this.getLimit() != null;
+    }
+
+    public Integer getLimit() {
+        return this.limit != null ? this.limit : (this.limit = super.getInt("limit"));
+    }
+
+    public Integer getOffset() {
+        return this.offset != null ? this.offset : (this.offset = super.getInt("offset"));
     }
 }
--- a/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
+++ b/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
@@ -1,0 +1,70 @@
+package com.microsoft.azure.cosmosdb.rx.internal.query;
+
+import com.microsoft.azure.cosmosdb.JsonSerializable;
+import com.microsoft.azure.cosmosdb.rx.internal.Utils;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OffsetContinuationToken extends JsonSerializable {
+    private static final String TOKEN_PROPERTY_NAME = "sourceToken";
+    private static final String OFFSET_PROPERTY_NAME = "offset";
+    private static final Logger logger = LoggerFactory.getLogger(CompositeContinuationToken.class);
+
+    public OffsetContinuationToken(int offset, String sourceToken) {
+
+        if (offset < 0) {
+            throw new IllegalArgumentException("offset should be non negative");
+        }
+
+        this.setOffset(offset);
+        this.setSourceToken(sourceToken);
+    }
+
+    public OffsetContinuationToken(String serializedCompositeToken) {
+        super(serializedCompositeToken);
+        this.getOffset();
+        this.getSourceToken();
+    }
+
+    private void setSourceToken(String sourceToken) {
+        super.set(TOKEN_PROPERTY_NAME, sourceToken);
+    }
+
+    private void setOffset(int offset) {
+        super.set(OFFSET_PROPERTY_NAME, offset);
+    }
+
+    public String getSourceToken() {
+        return super.getString(TOKEN_PROPERTY_NAME);
+    }
+
+    public int getOffset() {
+        if (super.getInt(OFFSET_PROPERTY_NAME) == null) {
+            return 0;
+        }
+        return super.getInt(OFFSET_PROPERTY_NAME);
+    }
+
+    public static boolean tryParse(String serializedOffsetContinuationToken,
+                                   Utils.ValueHolder<OffsetContinuationToken> outOffsetContinuationToken) {
+        if (StringUtils.isEmpty(serializedOffsetContinuationToken)) {
+            return false;
+        }
+
+        boolean parsed;
+        try {
+            outOffsetContinuationToken.v = new OffsetContinuationToken(serializedOffsetContinuationToken);
+            parsed = true;
+        } catch (Exception ex) {
+            logger.debug("Received exception {} when trying to parse: {}",
+                    ex.getMessage(),
+                    serializedOffsetContinuationToken);
+            parsed = false;
+            outOffsetContinuationToken.v = null;
+        }
+
+        return parsed;
+    }
+
+}
--- a/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
+++ b/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
@@ -6,7 +6,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class OffsetContinuationToken extends JsonSerializable {
+public final class OffsetContinuationToken extends JsonSerializable {
     private static final String TOKEN_PROPERTY_NAME = "sourceToken";
     private static final String OFFSET_PROPERTY_NAME = "offset";
     private static final Logger logger = LoggerFactory.getLogger(CompositeContinuationToken.class);
@@ -40,9 +40,6 @@ public class OffsetContinuationToken extends JsonSerializable {
     }
 
     public int getOffset() {
-        if (super.getInt(OFFSET_PROPERTY_NAME) == null) {
-            return 0;
-        }
         return super.getInt(OFFSET_PROPERTY_NAME);
     }
 
--- a/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
+++ b/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
@@ -61,7 +61,7 @@ public final class SkipDocumentQueryExecutionContext<T extends Resource> impleme
 
             Map<String, String> headers = new HashMap<>(tFeedResponse.getResponseHeaders());
             if (this.skipCount >= 0) {
-                // Add Take Continuation Token
+                // Add Offset Continuation Token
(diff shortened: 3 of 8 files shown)


sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/OffsetLimitQueryTests.java
```diff
@@ -0,0 +1,206 @@
...
+                logger.info(message, error);
+                throw new SkipException(message, error);
+            }
+            throw error;
+        }
+    }
+
+    @Test(groups = {"simple"}, timeOut = TIMEOUT)
+    public void offsetContinuationTokenRoundTrips() {
+        {
```
moderakh: redundant braces

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/TakeContinuationToken.java
```diff
@@ -75,6 +75,9 @@ public static boolean tryParse(String serializedTakeContinuationToken,
     }
 
     public int getTakeCount() {
+        if (super.getInt(LimitPropertyName) == null) {
+            return 0;
+        }
```
moderakh: you are deserializing twice.
You could store the value `super.getInt(LimitPropertyName)` in a variable and if not null return the value.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public String getSourceToken() {
+        return super.getString(TOKEN_PROPERTY_NAME);
+    }
+
+    public int getOffset() {
+        if (super.getInt(OFFSET_PROPERTY_NAME) == null) {
+            return 0;
+        }
+        return super.getInt(OFFSET_PROPERTY_NAME);
```
moderakh: you are deserilizaing twice.

You could store `super.getInt` in a variable and check null and return the variable

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/query/QueryInfo.java
```diff
@@ -41,6 +41,8 @@
     private Collection<AggregateOperator> aggregates;
     private Collection<String> orderByExpressions;
     private String rewrittenQuery;
+    private Integer offset;
```
bchong95: Can we make all of these final fields?

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
```diff
@@ -0,0 +1,70 @@
+package com.microsoft.azure.cosmosdb.rx.internal.query;
+
+import com.microsoft.azure.cosmosdb.JsonSerializable;
+import com.microsoft.azure.cosmosdb.rx.internal.Utils;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OffsetContinuationToken extends JsonSerializable {
```
bchong95: `final` class, since we don't want anyone inheriting from it.

bchong95: Make this private nested inside the corresponding execution component, since no other class needs to know about it. The goal is to write to an opaque string.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
```diff
@@ -0,0 +1,70 @@
...
+    private void setOffset(int offset) {
+        super.set(OFFSET_PROPERTY_NAME, offset);
+    }
+
+    public String getSourceToken() {
+        return super.getString(TOKEN_PROPERTY_NAME);
+    }
+
+    public int getOffset() {
+        if (super.getInt(OFFSET_PROPERTY_NAME) == null) {
```
bchong95: If the int is null then the token is probably malformed and you should throw an exception instead.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public final class SkipDocumentQueryExecutionContext<T extends Resource> implements IDocumentQueryExecutionComponent<T> {
+
+    private final IDocumentQueryExecutionComponent<T> component;
+    private int skipCount;
+
+    SkipDocumentQueryExecutionContext(IDocumentQueryExecutionComponent<T> component, int skipCount) {
```
bchong95: null check the source component.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
```diff
@@ -0,0 +1,80 @@
...
+        this.component = component;
+        this.skipCount = skipCount;
+    }
+
+    public static <T extends Resource> Observable<IDocumentQueryExecutionComponent<T>> createAsync(Function<String,
```
bchong95: nit: can these args be formatted a little?

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
```diff
@@ -0,0 +1,80 @@
...
+            Observable<IDocumentQueryExecutionComponent<T>>> createSourceComponentFunction, int skipCount,
+                                                                                                   String continuationToken) {
+        OffsetContinuationToken offsetContinuationToken;
+        Utils.ValueHolder<OffsetContinuationToken> outOffsetContinuationToken = new Utils.ValueHolder<>();
+
+        if (continuationToken != null) {
+            if (!OffsetContinuationToken.tryParse(continuationToken, outOffsetContinuationToken)) {
+                String message = String.format("Invalid JSON in continuation token %s for Parallel~Context",
```
bchong95: Parallel~Context -> Skip~Context

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/OffsetLimitQueryTests.java
```diff
@@ -0,0 +1,206 @@
...
+    @Test(groups = {"simple"}, timeOut = TIMEOUT * 10)
+    public void queryDocumentsWithTopContinuationTokens() {
+        int skipCount = 3;
+        int takeCount = 10;
+        String query = "SELECT * from c OFFSET " + skipCount + " LIMIT " + takeCount;
+        this.queryWithContinuationTokensAndPageSizes(query, new int[]{1, 5, 15}, takeCount);
+    }
+
+    private void queryWithContinuationTokensAndPageSizes(String query, int[] pageSizes, int takeCount) {
+        for (int pageSize : pageSizes) {
```
bchong95: add another loop to vary the maxItemCount.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/OffsetLimitQueryTests.java
mbhaskar: Removed

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/TakeContinuationToken.java
mbhaskar: Changed

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
mbhaskar: Changed implementation

gateway/src/main/java/com/microsoft/azure/cosmosdb/internal/query/QueryInfo.java
mbhaskar: Unable to make these final as we are lazy assigning values to these

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
mbhaskar: Yes propagating the exception now so that exception is thrown in tryParse

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
mbhaskar: Done

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
mbhaskar: Done

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/OffsetLimitQueryTests.java
mbhaskar: Page size is for maxItemCount

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/OffsetContinuationToken.java
mbhaskar: Made the class final. Leaving it as a separate class as all other continuation tokens are individual classes

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/SkipDocumentQueryExecutionContext.java
mbhaskar: Done

mbhaskar:
> Please add some samples in the examples project as later the customers will ask you for samples.
> 
> Is there any special support needed for partition split? If so if it is possible it is worth adding some tests covering partition split. (See DocumentProducerTest)

At the moment we have tests for this. Will add samples in seperate PR. Splits should work in the same way as any other query

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/OffsetLimitQueryTests.java
```diff
@@ -0,0 +1,210 @@
...
+    private List<Document> queryWithContinuationTokens(String query, int pageSize) {
+        String requestContinuation = null;
+        List<String> continuationTokens = new ArrayList<String>();
+        List<Document> receivedDocuments = new ArrayList<Document>();
+
+        do {
+            FeedOptions options = new FeedOptions();
+            options.setMaxItemCount(pageSize);
+            options.setEnableCrossPartitionQuery(true);
+            options.setMaxDegreeOfParallelism(2);
```
christopheranderson: nit: when you're not using the default values for something that's obviously under test, please either use descriptive variable names or leave comments next to literal usage so explain why you chose that value. In this case, why are we using cross partition and why are we using parallelism of 2? In theory, should we be testing a matrix of these values in e2e tests?

mbhaskar:
> Approved with suggestions. My nits aren't worth waiting for tests again, so you can treat them as advice instead.
> 
> Review checklist:
> 
> * [ ]  Dependencies - there are no dependencies added or modified
> * [ ]  Public API surface area - there are no changes to the public API surface area
> * [x]  Functionality - there are changes to the public functionality. This PR impacts query functionality. The SDK now supports SKIP/TAKE with continuation tokens.
> * [x]  Test coverage - this PR includes sufficient test coverage for the new functionality
> * Fundamentals:
> * [ ]  Perf - this should have no impact on perf
> * [ ]  Availability - this should have no impact on availability
> * [ ]  Consistency - this should have no impact on consistency
> * [ ]  Security - this should have no impact on security
> * [x]  Supportability - this will impact supportability because we added functionality. We should be sure to have samples that cover the new use cases. @mbhaskar - please create a tracking item to add this coverage.

#235

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/OffsetLimitQueryTests.java
mbhaskar: I just followed this pattern from tests for other query types. I will clean this up along with the example

PR #248, opened 2019-08-09
memory improvements (cherry pick of #230 to master)

moderakh:
This is the cherry pick of this PR (https://github.com/Azure/azure-cosmosdb-java/pull/230 for 2.4.6 hotfix release) onto master.

description of the cherry picked PR: https://github.com/Azure/azure-cosmosdb-java/pull/230
This PR Includes the following memory improvements.

(The following numbers are based on running benchmark tool on 16cpu core Linux VM with ReadThroughput workload and 4000 as degree of concurrency.)

1) when reading from network wire (through netty ByteBuf) we used to copy data once to ByteArrayOutputStream and from that we created String. Later ByteArrayOutputStream will get garbage collected, on high load this can lead to GCing 20-30% unnecessary data. The improvement is to pool the buffers.

2) we were deserializing replica physicalAddress in the AddressSelector to URI per request, we can cache the parsed URI form of the replica physical Address, also in the HttpTransportClient we were using URI for constructing paths which consumes more memory comparing to pure String. This 
somewhat contributed 5-10% memory which gets GCed.

3) we were decoding/encoding UUID to String unnecessarily. 2% memory improvement:

4) in auth token generation we are using StringBuilder which grows (it copies data on each grow) as we append to it. This is improved by preserving memory 1-2% memory improvement:

### DIFF
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/Uri.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/Uri.java
@@ -27,8 +27,8 @@ import java.net.URI;
 import java.util.Objects;
 
 public class Uri {
-    public String uriAsString;
-    public URI uri;
+    private final String uriAsString;
+    private final URI uri;
 
     public static Uri create(String uriAsString) {
         return new Uri(uriAsString);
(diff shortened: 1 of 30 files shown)


commons/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/Uri.java
```diff
@@ -0,0 +1,64 @@
...
+
+package com.microsoft.azure.cosmosdb.internal.directconnectivity;
+
+import java.net.URI;
+import java.util.Objects;
+
+public class Uri {
+    public String uriAsString;
+    public URI uri;
+
```
kushagraThapar: We should consider making them private and use public getters, so we keep the encapsulation intact.

christopheranderson:
Also needs manual port to azure-sdk-for-java (v3)

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/Uri.java
moderakh: added getters, the fields are final.

PR #256, opened 2019-09-04
offset-limit query fix

mbhaskar:
Fix for an issue where offset limit queries might not return expected number of results.


kirankumarkolli:
Aren't there test gaps which should be included.

mbhaskar:
> UT gaps should be immediate priority.

https://github.com/Azure/azure-cosmosdb-java/issues/257

kushagraThapar:
Force Merging the PR: NonValueAggregate is an independent issue and not a release blocker.

Issue #261, opened 2019-09-26
Cannot set MaxItemCount to -1 through Java SDK


avakutin:
**Describe the bug**
The documentation states that the maxItemCount property in the FeedOptions can be set to -1 to allow for the SDK to choose the optimal value, however, setting the maxItemCount to -1 returns a 400 with the message "One of the input values is invalid.: argumentName: MaxItemCount, message: Invalid MaxItemCount -1"

The stack trace points to a validation line in DocumentQueryExecutionContextFactory.createSpecializedDocumentQueryExecutionContextAsync:

BadRequestException validationError = Utils.checkRequestOrReturnException
(initialPageSize > 0, "MaxItemCount", "Invalid MaxItemCount %s", initialPageSize);

**To Reproduce**
FeedOptions options = new FeedOptions();
options.setMaxItemCount(-1);

Send cosmos request with the FeedOptions

**Expected behavior**
Should be able to set maxItemCount to -1 through the SDK

**Actual behavior**
BadRequestException{error=null, resourceAddress='null', statusCode=400, message=One of the input values is invalid.: argumentName: MaxItemCount, message: Invalid MaxItemCount -1, getCauseInfo=null, responseHeaders={}, requestHeaders=null}

**Environment summary**
Java JDK version: 2.6.1

kushagraThapar:
@mbhaskar this is a bug in the SDK. https://github.com/Azure/azure-cosmosdb-java/blob/master/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/DocumentQueryExecutionContextFactory.java#L183

We should also fix this in v3, and possibly v4.

kushagraThapar:
This has been released as v2.6.2, closing this issue.

PR #264, opened 2019-10-01
Removed BadRequestException when FeedOptions.maxItemCount is -1

kushagraThapar:
This PR fixes the issue of SDK throwing BadRequestException when maxItemCount is less than 0. 
https://github.com/Azure/azure-cosmosdb-java/issues/261


kushagraThapar:
> thanks for fixing it.
> You could avoid the outer "if", might be more readable this way.
> 
> ```java
> BadRequestException validationError = Utils.checkRequestOrReturnException(
>     initialPageSize > 0 || initialPageSize == -1, "MaxItemCount", "Invalid MaxItemCount %s", initialPageSize);
> if (validationError != null) {
>     return Observable.error(validationError);
> }
> ```

Yes, that seems good as well, I will do that.

kushagraThapar:
Timeout issues, merging the PR as the change is unrelated to the timeout issues.

PR #285, opened 2019-11-06
Adding collection routing map refresh flag on partition migration

simplynaveen20:
This PR will add x-ms-collectionroutingmap-refresh in header with value true in case of partition migration on backend , which will notify gateway to refresh collection routing map cache on address call

### DIFF
--- a/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
+++ b/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
@@ -288,6 +288,11 @@ public class GatewayAddressCache implements IAddressCache {
             headers.put(HttpConstants.HttpHeaders.FORCE_REFRESH, Boolean.TRUE.toString());
         }
 
+        if(request.forceCollectionRoutingMapRefresh)
+        {
+            headers.put(HttpConstants.HttpHeaders.FORCE_COLLECTION_ROUTING_MAP_REFRESH, Boolean.TRUE.toString());
+        }
+
         addressQuery.put(HttpConstants.QueryStrings.FILTER, HttpUtils.urlEncode(this.protocolFilter));
 
         addressQuery.put(HttpConstants.QueryStrings.PARTITION_KEY_RANGE_IDS, String.join(",", partitionKeyRangeIds));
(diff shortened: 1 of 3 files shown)


direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
```diff
@@ -288,6 +288,11 @@ private URL getServiceEndpoint() {
             headers.put(HttpConstants.HttpHeaders.FORCE_REFRESH, Boolean.TRUE.toString());
         }
 
+        if(request.forceCollectionRoutingMapRefresh)
+        {
```
moderakh: please fix code style on "{" here and elsewhere
opening "{" on the same line as if/else/while/try/catch

here and elsewhere.

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
```diff
@@ -483,6 +488,11 @@ public void dispose() {
             headers.put(HttpConstants.HttpHeaders.USE_MASTER_COLLECTION_RESOLVER, Boolean.TRUE.toString());
         }
 
+        if(request.forceCollectionRoutingMapRefresh)
+        {
```
moderakh: ditto

sdk/src/test/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCacheTest.java
```diff
@@ -120,20 +119,28 @@ public void getServerAddressesViaGateway(List<String> partitionKeyRangeIds,
                 null,
                 getCompositeHttpClient(configs));
 
-        RxDocumentServiceRequest req =
-                RxDocumentServiceRequest.create(OperationType.Create, ResourceType.Document,
-                        collectionLink + "/docs/",
-                getDocumentDefinition(), new HashMap<>());
+        for(int i = 0; i < 2 ;i++)
+        {
```
moderakh: ditto

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
simplynaveen20: Done

sdk/src/test/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCacheTest.java
simplynaveen20: Done

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
simplynaveen20: Done

Issue #298, opened 2019-12-14
[FEATURE] Add support for overriding default Direct TCP options


David-Noble-at-work:
When this issue is resolved you will be able to set Direct TCP options. In priority order we will take default Direct TCP options from:

1. The string value of system property `azure.cosmos.directTcp.defaultOptions`.
   Example: 
   ```
   -Dazure.cosmos.directTcp.defaultOptions={\"maxChannelsPerEndpoint\":5,\"maxRequestsPerChannel\":30}
   ```

2. The contents of the file located by system property `azure.cosmos.directTcp.defaultOptionsFile`.
   Example: 
   ```
   -Dazure.cosmos.directTcp.defaultOptionsFile=/path/to/default/options/file
   ```

3. The contents of the resource file named `azure.cosmos.directTcp.defaultOptions.json`.
   Specifically, the resource file is read from this stream: 
   ```java
   RntbdTransportClient.class.getClassLoader().getResourceAsStream("azure.cosmos.directTcp.defaultOptions.json")
   ```
   Example: Contents of resource file `azure.cosmos.directTcp.defaultOptions.json`.
   ```json
   {
     "bufferPageSize": 8192,
     "connectionTimeout": "PT1M",
     "idleChannelTimeout": "PT0S",
     "idleEndpointTimeout": "PT1M10S",
     "maxBufferCapacity": 8388608,
     "maxChannelsPerEndpoint": 10,
     "maxRequestsPerChannel": 30,
     "receiveHangDetectionTime": "PT1M5S",
     "requestExpiryInterval": "PT5S",
     "requestTimeout": "PT1M",
     "requestTimerResolution": "PT0.5S",
     "sendHangDetectionTime": "PT10S",
     "shutdownTimeout": "PT15S"
   }

4. Otherwise, if none of these values are set or all set values are in error we will create default options based on a set of hard-wired values defined in the default private parameterless constructor for `RntbdTransportClient.Options`.

David-Noble-at-work:
Work completed. See PR#308.

PR #299, opened 2019-12-16
Add support for specifying default Direct TCP options using System properties

David-Noble-at-work:
See #298. Also: 
* Added the ability to set the Direct TCP protocol using:
  - `-Dcosmos.directModeProtocol`\
     the current mechanism
  - `-Dazure.cosmos.directModeProtocl`\
     for consistency with v4 and azure sdk guidelines
  - `DIRECT_MODE_PROTOCOL` environment variable\
    for consistency with other settings that we vary in CI test runs.

### DIFF
--- a/direct-impl/pom.xml
+++ b/direct-impl/pom.xml
@@ -36,6 +36,7 @@ SOFTWARE.
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <test.groups>unit</test.groups>
     <guava.version>27.0.1-jre</guava.version>
+    <jackson-datatype.version>2.9.9</jackson-datatype.version>
     <metrics.version>4.1.0</metrics.version>
     <micrometer.version>1.2.0</micrometer.version>
   </properties>
@@ -294,6 +295,11 @@ SOFTWARE.
       <artifactId>azure-cosmosdb-commons-test-utils</artifactId>
       <version>${project.version}</version>
       <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+      <version>${jackson-datatype.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
--- a/direct-impl/pom.xml
+++ b/direct-impl/pom.xml
@@ -36,7 +36,7 @@ SOFTWARE.
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <test.groups>unit</test.groups>
     <guava.version>27.0.1-jre</guava.version>
-    <jackson-datatype.version>2.9.9</jackson-datatype.version>
+    <jackson-datatype-jsr310.version>2.9.9</jackson-datatype-jsr310.version>
     <metrics.version>4.1.0</metrics.version>
     <micrometer.version>1.2.0</micrometer.version>
   </properties>
@@ -299,7 +299,7 @@ SOFTWARE.
     <dependency>
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jsr310</artifactId>
-      <version>${jackson-datatype.version}</version>
+      <version>${jackson-datatype-jsr310.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
--- a/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
+++ b/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
@@ -225,6 +225,9 @@ public final class RntbdTransportClient extends TransportClient implements AutoC
         private final Duration requestTimeout;
 
         @JsonProperty()
+        private final Duration requestTimerResolution;
+
+        @JsonProperty()
         private final Duration sendHangDetectionTime;
 
         @JsonProperty()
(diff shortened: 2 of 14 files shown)


direct-impl/pom.xml
```diff
@@ -36,6 +36,7 @@ SOFTWARE.
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <test.groups>unit</test.groups>
     <guava.version>27.0.1-jre</guava.version>
+    <jackson-datatype.version>2.9.9</jackson-datatype.version>
```
moderakh: please don't overwrite jackson version here, the default dependency should come from a common place. that should be either the parent project or the commons subproject.

overwriting here may cause multiple versions of the dependencies ....

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
```diff
@@ -310,35 +343,100 @@ public UserAgentContainer userAgent() {
...
-            private Duration idleEndpointTimeout = SEVENTY_SECONDS;
-            private int maxBufferCapacity = 8192 << 10;
-            private int maxChannelsPerEndpoint = 10;
-            private int maxRequestsPerChannel = 30;
-            private int partitionCount = 1;
-            private Duration receiveHangDetectionTime = SIXTY_FIVE_SECONDS;
-            private Duration requestExpiryInterval = FIVE_SECONDS;
+            private static final Options DEFAULT_OPTIONS;
+
+            static {
```
moderakh: as this is in the static initializer section and we are doing IO operations inside,
if any uncaught failure happen in the static initializer the class cannot load.

I am thinking we probably should have higher level catch to catch any java.lang.Error in case something fails. This is my take from the similar failure in the static initializer in the hybridrow which faced for spark connector a while back.

David-Noble-at-work: I agree. let me add that to completely avoid the issue of exceptions during initialization.

David-Noble-at-work: I'm going to wrap all but the final line in a try block. In the finally block there will be just one thing that can fail: allocation of `new Options()`, should the try block not produce options (due to failure or--more likely--because no default options have been specified.

direct-impl/pom.xml
David-Noble-at-work: good catch. I meant to remove this.

David-Noble-at-work: This isn't the version property I thought it was. I am including this additional dependency--which is independently of Jackson proper--to pick up `Duration` deserialization on Java 8 for versions of the Jackson library below 3.0. I could update the version property name to make this a little more clear:
```
    <jackson-datatype-jsr310.version>2.9.9</jackson-datatype-jsr310.version>
...
    <dependency>
      <groupId>com.fasterxml.jackson.datatype</groupId>
      <artifactId>jackson-datatype-jsr310</artifactId>
      <version>${jackson-datatype-jsr310.version}</version>
    </dependency>
```
2.9.9 is the version closest to the version we're using: 2.9.9.3.

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
```diff
@@ -310,35 +343,103 @@ public UserAgentContainer userAgent() {
...
+                            } catch (IOException error) {
+                                logger.error("failed to load default Direct TCP options from {} due to ", path, error);
+                            }
+                        }
+                    }
+
+                    if (options == null) {
+
+                        String name = propertyName + ".json";
+                        InputStream stream = RntbdTransportClient.class.getClassLoader().getResourceAsStream(name);
```
moderakh: in case of a failure something needs to close the `InputStream`

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
```diff
@@ -310,35 +343,103 @@ public UserAgentContainer userAgent() {
...
+                        try {
+                            options = RntbdObjectMapper.readValue(string, Options.class);
+                        } catch (IOException error) {
+                            logger.error("failed to parse default Direct TCP options {} due to ", string, error);
+                        }
+                    }
+
+                    if (options == null) {
+
+                        String path = System.getProperty(propertyName + "File");
```
moderakh: This will evaluate to `azure.cosmos.directTcp.defaultOptionsFile`  right?

I would make this a constant rather than building it from String concatenation. "azure.cosmos.directTcp.defaultOptions" + "File"

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
```diff
@@ -310,35 +343,103 @@ public UserAgentContainer userAgent() {
...
+                //
+                // Otherwise, if none of these values are set or an error occurs we create default options based on a
+                // set of hard-wired values defined in the default private parameterless constructor for
+                // RntbdTransportClient.Options.
+
+                Options options = null;
+
+                try {
+                    String propertyName = "azure.cosmos.directTcp.defaultOptions";
+                    String string = System.getProperty(propertyName);
```
moderakh: "azure.cosmos.directTcp.defaultOptions" should be defined as a constant.
Maybe Configs.java would be the right place for this. As the rest of the property configs are defined there.

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
David-Noble-at-work: why?

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
David-Noble-at-work: sure

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/RntbdTransportClient.java
David-Noble-at-work: this observation by itself made the code review worthwhile. thanks for catching this bug.

PR #315, opened 2020-01-31
move deserialization out of netty IO thread

moderakh:
move deserialization out of netty IO thread


direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/TransportClient.java
```diff
@@ -24,15 +24,18 @@
...
 
 import java.net.URI;
 
 public abstract class TransportClient implements AutoCloseable {
 
     // Uses requests's ResourceOperation to determine the operation
     public Single<StoreResponse> invokeResourceOperationAsync(Uri physicalAddress, RxDocumentServiceRequest request) {
-        return this.invokeStoreAsync(physicalAddress, new ResourceOperation(request.getOperationType(), request.getResourceType()), request);
+        return this.invokeStoreAsync(physicalAddress, new ResourceOperation(request.getOperationType(), request.getResourceType()), request)
+                .observeOn(Schedulers.computation());
```
kushagraThapar: Why do we need to change the thread in this method ?
Where is the deserialization happening ?

moderakh: we free up the netty thread right after it returns the result. There are a lot of computation in the consistency stack which will be happening after this including deserialization. for the same reason I am moving everything off the netty thread.

kushagraThapar: I see, okay that makes sense.

PR #332, opened 2020-05-08
Fixing an issue with order by query continuation token results

mbhaskar:
This PR fixes an issue due to which the order by results when resuming from continuation token might contain duplicates/missing documents
- Adding a new method in ResourceId
- Adding a new test class to test the above scenario. This class can eventually be enhanced to add more tests around query validation.

### DIFF
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ResourceId.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ResourceId.java
@@ -381,17 +381,13 @@ public class ResourceId {
         return rid;
     }
 
-    public long getDocument() {
-        return this.document;
-    }
-
     /**
-     * ResourceID#getDocument returns the long value of the document in Big Endian, this method reverses the bytes 
+     * Returns the long value of the document. The value computed is in Big Endian, so this method reverses the bytes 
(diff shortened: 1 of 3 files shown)


commons/src/main/java/com/microsoft/azure/cosmosdb/internal/ResourceId.java
```diff
@@ -385,14 +385,24 @@ public long getDocument() {
...
     }
 
+    /**
+     * ResourceID#getDocument returns the long value of the document in Big Endian, this method reverses the bytes 
+     * and returns Little Endian order value of the long
+     * 
+     * @return document long value
+     */
+    public long getDocumentLittleEndian() {
+        return Long.reverseBytes(this.document);
```
moderakh, line 395: we should change the getDocument() to return the correct value instead of adding a new method

mbhaskar: Ok, didnt want to effect anything that relies on existing value. Let me try changing that and push to see what the CI does.

mbhaskar: Removed the new method and renamed.

PR #341, opened 2020-06-15
Fix adobe duplicate result using continuation token

kushagraThapar:
* This PR is a duplicate of PR : https://github.com/Azure/azure-cosmosdb-java/pull/339
* Since that PR is experiencing some issues with running CI because of permissions.


sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/QueryValidationTests.java
```diff
@@ -164,4 +212,27 @@ public String getCollectionLink() {
...
     }
 
+    private void queryWithOrderByAndAssert(
+            int pageSize,
+            int documentCount,
+            String query,
+            String orderByPropName,
+            String collectionLink,
+            List<Document> documentsInserted) {
+        List<Document> documentsPaged = queryWithContinuationTokens(query, pageSize, collectionLink);
```
moderakh, line 222: why are we still dealing with Document in v4? why aren't we testing on the public apis?

PR #344, opened 2020-06-23
Fix Method Name Typo in Murmurhash

southpolesteve:


commons/src/main/java/com/microsoft/azure/cosmosdb/internal/routing/MurmurHash3_128.java
```diff
@@ -150,7 +150,7 @@ private static final long getLittleEndianLong(byte[] bytes, int offset) {
                 | ((bytes[offset] & 0xffL));
     }
 
-    private static int intAsLittleIndian(byte[] bytes, int i) {
+    private static int intAsLittleEndian(byte[] bytes, int i) {
```
moderakh, line 153: funny typo! thanks for fixing it.

PR #350, opened 2020-08-09
[BugFix] Fix Header casing of DocumentServiceResponse to be case insensitve header comparison

visridha:
Http Headers are by default case insensitive. Fix DocumentServiceResponse so that the headers coming out of it are case insensitive.

### DIFF
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/RxDocumentServiceResponse.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/RxDocumentServiceResponse.java
@@ -52,10 +52,7 @@ import org.apache.commons.lang3.StringUtils;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * This is core Transport/Connection agnostic response for the Azure Cosmos DB database service.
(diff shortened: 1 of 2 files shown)


commons/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/RxDocumentServiceResponse.java
```diff
@@ -69,7 +66,8 @@ public RxDocumentServiceResponse(StoreResponse response) {
         String[] headerNames = response.getResponseHeaderNames();
         String[] headerValues = response.getResponseHeaderValues();
 
-        this.headersMap = new HashMap<>(headerNames.length);
+        // headers in CosmosDB are case insensitive.
+        this.headersMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
moderakh, line 70: Thanks Vinod. although as per http spec header keys are case sensitive, DSR is also used for RNTBD and we haven't seen any issue there. 

What's the context? have you seen a relevant issue?

commons/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/RxDocumentServiceResponse.java
```diff
@@ -52,10 +52,7 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
moderakh, line 55: as per code style we don't use wildcard import.

commons/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/RxDocumentServiceResponse.java
visridha: This was seen in Gateway requests when trying against Compute (which changes the header casing due to optimizations in the C# Http layer). We made it so that it's backcompat with the casing but we want to not do the additional work going forward. Once this fix releases we can remove that compat for newer SDK versions...

commons/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/RxDocumentServiceResponse.java
visridha: Fixed.

kirankumarkolli:
@moderakh are there any other headers implementations inside which needs attenton?

PR #351, opened 2020-08-26
Fixing region fail over issue for direct tcp calls

simplynaveen20:
Currently for tcp network exception we are doing region failover, which is causing problem for some users where due to some intermittent network issue on BE nodes sdk is sending traffic for next 5 min to other available region.

This PR will limit the region failover only in case of http network failure.

### DIFF
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
@@ -52,7 +52,7 @@ public class DocumentClientException extends Exception {
 
     private final Map<String, String> requestHeaders;
     private final int statusCode;
-    private volatile Map<String, String> responseHeaders;
+    private Map<String, String> responseHeaders;
 
     private volatile ClientSideRequestStatistics clientSideRequestStatistics;
     private volatile Error error;
@@ -322,10 +322,6 @@ public class DocumentClientException extends Exception {
     }
 
 	void setSubStatusCode(int subStatusCode) {
-		if (this.responseHeaders == null) {
-			this.responseHeaders = new HashMap<>();
-		}
-
 		this.responseHeaders.put(HttpConstants.HttpHeaders.SUB_STATUS, Integer.toString(subStatusCode));
 	}
 
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
@@ -51,7 +51,7 @@ public class DocumentClientException extends Exception {
     private static final long serialVersionUID = 1L;
 
     private final Map<String, String> requestHeaders;
-	private final int statusCode;
+    private final int statusCode;
     private volatile Map<String, String> responseHeaders;
 
     private volatile ClientSideRequestStatistics clientSideRequestStatistics;
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
@@ -52,7 +52,7 @@ public class DocumentClientException extends Exception {
 
     private final Map<String, String> requestHeaders;
     private final int statusCode;
-    private Map<String, String> responseHeaders;
+    private final Map<String, String> responseHeaders;
 
     private volatile ClientSideRequestStatistics clientSideRequestStatistics;
     private volatile Error error;
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
@@ -28,9 +28,9 @@ import com.microsoft.azure.cosmosdb.internal.HttpConstants;
 import com.microsoft.azure.cosmosdb.internal.directconnectivity.Uri;
 import org.apache.commons.lang3.StringUtils;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
 /**
  * This class defines a custom exception type for all operations on
@@ -51,8 +51,8 @@ public class DocumentClientException extends Exception {
     private static final long serialVersionUID = 1L;
 
     private final Map<String, String> requestHeaders;
-    private final Map<String, String> responseHeaders;
-    private final int statusCode;
+	private final int statusCode;
+    private volatile Map<String, String> responseHeaders;
 
     private volatile ClientSideRequestStatistics clientSideRequestStatistics;
     private volatile Error error;
@@ -320,6 +320,14 @@ public class DocumentClientException extends Exception {
                 ", requestHeaders=" + requestHeaders +
                 '}';
     }
+
+	void setSubStatusCode(int subStatusCode) {
+		if (this.responseHeaders == null) {
+			this.responseHeaders = new HashMap<>();
+		}
+
+		this.responseHeaders.put(HttpConstants.HttpHeaders.SUB_STATUS, Integer.toString(subStatusCode));
+	}
 
     String getInnerErrorMessage() {
         String innerErrorMessage = super.getMessage();
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/HttpConstants.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/internal/HttpConstants.java
@@ -323,6 +323,9 @@ public class HttpConstants {
 
         // 404: LSN in session token is higher
         public static final int READ_SESSION_NOT_AVAILABLE = 1002;
+
+        // Client generated gateway network error substatus
+        public static final int GATEWAY_ENDPOINT_UNAVAILABLE = 10001;
     }
 
     public static class HeaderValues {
--- a/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
+++ b/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
@@ -333,14 +333,37 @@ public class GatewayAddressCache implements IAddressCache {
         Single<RxDocumentServiceResponse> dsrObs = responseObs.toSingle().flatMap(rsp ->
                 HttpClientUtils.parseResponseAsync(rsp));
         return dsrObs.map(
-                dsr -> {
-                    if (logger.isDebugEnabled()) {
-                        logger.debug("getServerAddressesViaGatewayAsync deserializes result");
-                    }
-                    logAddressResolutionEnd(request, identifier);
-                    List<Address> addresses = dsr.getQueryResponse(Address.class);
-                    return addresses;
-                });
+        dsr -> {
+            if (logger.isDebugEnabled()) {
+                logger.debug("getServerAddressesViaGatewayAsync deserializes result");
+            }
+            logAddressResolutionEnd(request, identifier);
+            List<Address> addresses = dsr.getQueryResponse(Address.class);
+            return addresses;
+        }).onErrorResumeNext(throwable -> {
+            if (!(throwable instanceof Exception)) {
+                // fatal error
+                logger.error("Unexpected failure {}", throwable.getMessage(), throwable);
+                return Single.error(throwable);
+            }
+
+            Exception exception = (Exception) throwable;
+            DocumentClientException dce;
+            if (!(exception instanceof DocumentClientException)) {
+                // wrap in DocumentClientException
+                logger.error("Network failure", exception);
+                dce = new DocumentClientException(0, exception);
+                BridgeInternal.setRequestHeaders(dce, request.getHeaders());
+            } else {
+                dce = (DocumentClientException) exception;
+            }
+
+            if (WebExceptionUtility.isNetworkFailure(dce)) {
+                BridgeInternal.setSubStatusCode(dce, HttpConstants.SubStatusCodes.GATEWAY_ENDPOINT_UNAVAILABLE);
+            }
+
+            return Single.error(dce);
+        });
     }
 
     public void dispose() {
--- a/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
+++ b/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
@@ -127,16 +127,16 @@ public class ClientRetryPolicy implements IDocumentClientRetryPolicy {
 
         // Received Connection error (HttpRequestException), initiate the endpoint rediscovery
         if (WebExceptionUtility.isNetworkFailure(e)) {
-            if (this.isReadRequest || WebExceptionUtility.isWebExceptionRetriable(e)) {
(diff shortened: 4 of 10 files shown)


gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -126,8 +128,13 @@ public ClientRetryPolicy(GlobalEndpointManager globalEndpointManager,
         // Received Connection error (HttpRequestException), initiate the endpoint rediscovery
         if (WebExceptionUtility.isNetworkFailure(e)) {
             if (this.isReadRequest || WebExceptionUtility.isWebExceptionRetriable(e)) {
-                logger.warn("Endpoint not reachable. Will refresh cache and retry. ", e);
-                return this.shouldRetryOnEndpointFailureAsync(this.isReadRequest, false);
+                if (clientException != null && Exceptions.isSubStatusCode(clientException, HttpConstants.SubStatusCodes.GATEWAY_ENDPOINT_UNAVAILABLE)) {
```
simplynaveen20, line 131: Will do similar check for write in next iteration

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -63,6 +64,7 @@
     private RetryContext retryContext;
     private ClientSideRequestStatistics clientSideRequestStatistics;
     private AtomicInteger cnt = new AtomicInteger(0);
+    private int serviceUnavailableRetryCount;
```
kushagraThapar, line 67: Shouldn't this be volatile ?
Please check other retry policies code to make sure we are covering the race conditions.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -207,6 +214,36 @@ private ShouldRetryResult shouldRetryOnSessionNotAvailable() {
         return refreshCompletion.andThen(Single.just(ShouldRetryResult.noRetry()));
     }
 
+    private Single<ShouldRetryResult> shouldRetryOnBackendServiceUnavailableAsync(boolean isReadRequest, boolean isWebExceptionRetriable) {
```
kushagraThapar, line 217: So this method will only be called when the status code is not `10001` ?
which would mean for all other sub status codes ?

simplynaveen20: 10001 is now new substatus code which will be there in all gateway request if they face  network error. For others we dont want to do failover

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
simplynaveen20: this should be fine as we already have failoverRetryCount in this class as non volatile

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
```diff
@@ -333,14 +333,37 @@ private URL getServiceEndpoint() {
...
+                logger.error("Unexpected failure {}", throwable.getMessage(), throwable);
+                return Single.error(throwable);
+            }
+
+            Exception exception = (Exception) throwable;
+            DocumentClientException dce;
+            if (!(exception instanceof DocumentClientException)) {
+                // wrap in DocumentClientException
+                logger.error("Network failure", exception);
+                dce = new DocumentClientException(0, exception);
```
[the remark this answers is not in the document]
simplynaveen20: We don't have any constructor for exception alone, and this code is carried over from RxGatewayStoreModel which is main class for gateway calls. Ideally address refresh should go through same stack , but it is not in scope of this PR.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
kushagraThapar: Could that be a bug too, we should verify these race conditions going forward. 
@moderakh - what are your thoughts on `volatile` / `AtomicInteger` / `non-volatile` counter.

direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/GatewayAddressCache.java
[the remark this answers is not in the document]
kirankumarkolli: "0" status code will byte back. Lets revisit it.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/CollectionCrudTest.java
```diff
@@ -255,7 +255,7 @@ public void replaceCollection(String collectionName, boolean isNameBased)  {
 
         // validate
         ResourceResponseValidator<DocumentCollection> validator = new ResourceResponseValidator.Builder<DocumentCollection>()
-                .indexingMode(IndexingMode.Lazy).build();
+                .indexingMode(IndexingMode.Consistent).build();
```
kirankumarkolli, line 258: Best if it can be a separate PR

simplynaveen20: Will revert this in next iteration, below is the PR
https://github.com/Azure/azure-cosmosdb-java/pull/354

kushagraThapar:
Discussed the above comments with @simplynaveen20 offline. Good to go from my side.

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
```diff
@@ -51,8 +51,8 @@
     private static final long serialVersionUID = 1L;
 
     private final Map<String, String> requestHeaders;
-    private final Map<String, String> responseHeaders;
     private final int statusCode;
+    private volatile Map<String, String> responseHeaders;
```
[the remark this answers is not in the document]
simplynaveen20: We have to remove final to set  subStatuscode , and constructor are getting used at some place with null response header.

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
```diff
@@ -321,6 +321,14 @@ public String toString() {
                 '}';
     }
 
+	void setSubStatusCode(int subStatusCode) {
+		if (this.responseHeaders == null) {
```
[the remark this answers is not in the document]
simplynaveen20: constructor are getting used at some place with null response header.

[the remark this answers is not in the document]
moderakh: all the constructors at the end of the day will invoke the following one:

```
    private DocumentClientException(
        int statusCode, String message, Map<String, String> responseHeaders, Throwable cause) {
        super(message, cause);
        this.requestHeaders = new ConcurrentHashMap<>();
        this.responseHeaders = responseHeaders == null
            ? new ConcurrentHashMap<>()
            : new ConcurrentHashMap<>(responseHeaders);
        this.statusCode = statusCode;
    }
```

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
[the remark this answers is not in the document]
moderakh: all the constructors are invoking this one which initializes the headers to a non null value. we should stick to the existing constructor and not re-initialize responseHeaders outside of the constructor

https://github.com/Azure/azure-cosmosdb-java/blob/master/commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java#L141-L149

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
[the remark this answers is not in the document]
simplynaveen20: done

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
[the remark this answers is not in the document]
simplynaveen20: done

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -207,6 +214,36 @@ private ShouldRetryResult shouldRetryOnSessionNotAvailable() {
...
+        if (!isReadRequest && !isWebExceptionRetriable) {
+            logger.warn("shouldRetryOnBackendServiceUnavailableAsync() Not retrying on write with non retriable exception. Retry count = {}", this.serviceUnavailableRetryCount);
+            return Single.just(ShouldRetryResult.noRetry());
+        }
+
+        if (this.serviceUnavailableRetryCount++ > MaxServiceUnavailableRetryCount) {
+			logger.warn("shouldRetryOnBackendServiceUnavailableAsync() Not retrying. Retry count = {}", this.serviceUnavailableRetryCount);
```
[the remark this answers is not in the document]
simplynaveen20: done

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
```diff
@@ -51,8 +51,8 @@
     private static final long serialVersionUID = 1L;
 
     private final Map<String, String> requestHeaders;
-    private final Map<String, String> responseHeaders;
     private final int statusCode;
+    private Map<String, String> responseHeaders;
```
moderakh, line 55: please change to final.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -207,6 +214,36 @@ private ShouldRetryResult shouldRetryOnSessionNotAvailable() {
...
+            logger.warn("shouldRetryOnBackendServiceUnavailableAsync() Not retrying on write with non retriable exception. Retry count = {}", this.serviceUnavailableRetryCount);
+            return Single.just(ShouldRetryResult.noRetry());
+        }
+
+        if (this.serviceUnavailableRetryCount++ > MaxServiceUnavailableRetryCount) {
+            logger.warn("shouldRetryOnBackendServiceUnavailableAsync() Not retrying. Retry count = {}", this.serviceUnavailableRetryCount);
+            return Single.just(ShouldRetryResult.noRetry());
+        }
+
+		if (!this.canUseMultipleWriteLocations && ! isReadRequest) {
```
moderakh, line 228: invalid indentation.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -207,6 +214,36 @@ private ShouldRetryResult shouldRetryOnSessionNotAvailable() {
...
+			// Write requests on single master cannot be retried, no other regions available
+			return Single.just(ShouldRetryResult.noRetry());
+		}
+
+		int availablePreferredLocations = this.globalEndpointManager.getPreferredLocationCount();
+		if (availablePreferredLocations <= 1) {
+			logger.warn("shouldRetryOnServiceUnavailable() Not retrying. No other regions available for the request. AvailablePreferredLocations = {}", availablePreferredLocations);
```
moderakh, line 235: invalid indentation

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
moderakh, line 67: serviceUnavailableRetryCount is not set by anything in your code. 

Is some logic missing? or do we need this at all?

moderakh: serviceUnavailableRetryCount is not set by anything in your code.

Is some logic missing? or do we need this at all?

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
```diff
@@ -207,6 +214,36 @@ private ShouldRetryResult shouldRetryOnSessionNotAvailable() {
         return refreshCompletion.andThen(Single.just(ShouldRetryResult.noRetry()));
     }
 
+    private Single<ShouldRetryResult> shouldRetryOnBackendServiceUnavailableAsync(boolean isReadRequest, boolean isWebExceptionRetriable) {
+        if (!isReadRequest && !isWebExceptionRetriable) {
+            logger.warn("shouldRetryOnBackendServiceUnavailableAsync() Not retrying on write with non retriable exception. Retry count = {}", this.serviceUnavailableRetryCount);
+            return Single.just(ShouldRetryResult.noRetry());
+        }
+
+        if (this.serviceUnavailableRetryCount++ > MaxServiceUnavailableRetryCount) {
```
moderakh, line 223: serviceUnavailableRetryCount value always be zero. you are not setting/incrementing it anywhere.

gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/ClientRetryPolicy.java
moderakh: volatile is fine. as the retry although happening on different threads will not happen concurrently.

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
simplynaveen20: because of set sub status method, it cant be final

simplynaveen20:
> looks good to me thanks Naveen.
> 
> Once scenario we haven't covered is when address resolution request to GW fails.
> This will cause a region failover, but instead we should be more lenient toward GW address all and not fail over the region.

Yes as discussed , its not only address resolution, we need to think and change region fail over for all GW calls.

commons/src/main/java/com/microsoft/azure/cosmosdb/DocumentClientException.java
moderakh: you are not initializing responseHeaders in setSubstatus method.

So it should be possible to make this final.

simplynaveen20: sorry my mistake i thought it otherway , thanks

commons/src/main/java/com/microsoft/azure/cosmosdb/internal/HttpConstants.java
```diff
@@ -323,6 +323,9 @@
 
         // 404: LSN in session token is higher
         public static final int READ_SESSION_NOT_AVAILABLE = 1002;
+
+        // Client generated gateway network error substatus
+        public static final int GATEWAY_ENDPOINT_UNAVAILABLE = 10001;
```
moderakh, line 328: we should use a substatus code which doesn't collide with the service side substatus codes otherwise the client-retry-policy will be confused.

PR #356, opened 2020-09-01
Updating version to 2.6.11-SNAPSHOT and updating Readme and change log

simplynaveen20:
Updating version to 2.6.11-SNAPSHOT and updating Readme and change log


changelog/README.md
```diff
@@ -1,3 +1,6 @@
+## 2.6.10 - 2020-08-31
+- Fixes region fail over issue for direct tcp calls ([#351](https://github.com/Azure/azure-cosmosdb-java/pull/351))
```
kirankumarkolli, line 2: Lets please expand with more context around what CX sees and the expected new behavior.

simplynaveen20: added more context, check the text now

Issue #359, opened 2020-09-11
QueryPlan should be a read request


simplynaveen20:
**Describe the bug**
QueryPlan should be a read request, instead currently it is a write request in sdk

**Expected behavior**
QueryPlan should be read request

**Actual behavior**
QueryPlan is write request

**Environment summary**
SDK Version: V2 2.6.10

**Additional context**
This fix need to ported in V3 and V4 sdk as well

PR #360, opened 2020-09-11
Improving http request time out avoiding region failover for high latency request

simplynaveen20:
Currently we have an api on ConnectionPolicy setRequestTimeoutInMillis which sets the request time out for both TCP and HTTP requests.
We want to cap the lower value of http client request time out to 60 sec, as http network request call ends up into region fail over and also Gateway does not have any sla.

**Note:** This PR also contains the improvement on AddressResolutionStatistics in ClientSideRequestStatistics, which include error message if there is any failure on address cache refresh call, also including flag inflightRequest which will indicate whether address call finished in lifetime of the the request.

**Also closes query plan issue**
closes https://github.com/Azure/azure-cosmosdb-java/issues/359

### DIFF
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
@@ -123,7 +123,7 @@ public class ClientSideRequestStatistics {
         AddressResolutionStatistics resolutionStatistics = new AddressResolutionStatistics();
         resolutionStatistics.startTime = ZonedDateTime.now(ZoneOffset.UTC);
         //  Very far in the future
-        resolutionStatistics.endTime = ZonedDateTime.of(LocalDateTime.MAX, ZoneOffset.UTC);
+        resolutionStatistics.endTime = null;
         resolutionStatistics.targetEndpoint = targetEndpoint == null ? "<NULL>" : targetEndpoint.toString();
 
         synchronized (this) {
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
@@ -22,6 +22,13 @@
  */
 package com.microsoft.azure.cosmosdb;
 
+import com.microsoft.azure.cosmosdb.internal.OperationType;
+import com.microsoft.azure.cosmosdb.internal.ResourceType;
+import com.microsoft.azure.cosmosdb.internal.Utils;
+import com.microsoft.azure.cosmosdb.internal.directconnectivity.StoreResult;
+import com.microsoft.azure.cosmosdb.rx.internal.RxDocumentServiceRequest;
+import org.apache.commons.lang3.StringUtils;
+
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Duration;
@@ -36,13 +43,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
-
-import com.microsoft.azure.cosmosdb.internal.OperationType;
-import com.microsoft.azure.cosmosdb.internal.ResourceType;
-import com.microsoft.azure.cosmosdb.internal.Utils;
-import com.microsoft.azure.cosmosdb.internal.directconnectivity.StoreResult;
-import com.microsoft.azure.cosmosdb.rx.internal.RxDocumentServiceRequest;
-import org.apache.commons.lang3.StringUtils;
 
 public class ClientSideRequestStatistics {
 
@@ -133,7 +133,7 @@ public class ClientSideRequestStatistics {
         return identifier;
     }
 
-    public void recordAddressResolutionEnd(String identifier) {
+    public void recordAddressResolutionEnd(String identifier, String errorMessage) {
         if (StringUtils.isEmpty(identifier)) {
             return;
         }
@@ -150,6 +150,8 @@ public class ClientSideRequestStatistics {
 
             AddressResolutionStatistics resolutionStatistics = this.addressResolutionStatistics.get(identifier);
             resolutionStatistics.endTime = responseTime;
+            resolutionStatistics.errorMessage = errorMessage;
+            resolutionStatistics.inflightRequest = false;
         }
     }
 
@@ -259,6 +261,8 @@ public class ClientSideRequestStatistics {
         private ZonedDateTime startTime;
         private ZonedDateTime endTime;
         private String targetEndpoint;
+        private String errorMessage;
+        private boolean inflightRequest = true;
 
         AddressResolutionStatistics() {
         }
@@ -266,10 +270,12 @@ public class ClientSideRequestStatistics {
         @Override
         public String toString() {
             return "AddressResolutionStatistics{" +
-                    "startTime=\"" + formatDateTime(startTime) + "\"" +
-                    ", endTime=\"" + formatDateTime(endTime) + "\"" +
-                    ", targetEndpoint='" + targetEndpoint + '\'' +
-                    '}';
+            "startTime=\"" + formatDateTime(startTime) + "\"" +
+            ", endTime=\"" + formatDateTime(endTime) + "\"" +
+            ", inflightRequest='" + inflightRequest + '\'' +
+            ", targetEndpoint='" + targetEndpoint + '\'' +
+            ", errorMessage='" + errorMessage + '\'' +
+            '}';
         }
     }
 }
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
@@ -32,7 +32,6 @@ import org.apache.commons.lang3.StringUtils;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Duration;
-import java.time.LocalDateTime;
 import java.time.ZoneOffset;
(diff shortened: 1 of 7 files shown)


sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
```diff
@@ -0,0 +1,254 @@
...
+			// Scenario before restricting http readTimeout i.e. verifying failure by updating requestTimeoutInMillis via reflection
+			// we should see the request failure and region fail over
+			httpClientFactory.withRequestTimeoutInMillis(1);
+			FieldUtils.writeField(httpClientFactory, "requestTimeoutInMillis", 1, true);
+			httpClient = httpClientFactory.toHttpClientBuilder().build();
+			FieldUtils.writeField(storeModel, "httpClient", httpClient, true);
+			queryObservable = client.queryDocuments(SHARED_MULTI_PARTITION_COLLECTION.getSelfLink(), "Select * from C", feedOptions);
+			FailureValidator failureValidator = new FailureValidator.Builder().causeInstanceOf(ClosedChannelException.class).build();
+			validateQueryFailure(queryObservable, failureValidator);
+			// TODO https://github.com/Azure/azure-cosmosdb-java/issues/359
```
simplynaveen20, line 243: Should we including this fix too , making query plan read call ? Its just a small change and possibly no impact , but I am little hesitant touching anything out of current issue context

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
```diff
@@ -0,0 +1,254 @@
...
+			fail("Client initialization should not fail due to low requestTimeout on connection policy " + ex.getMessage());
+		} finally {
+			if (client != null) {
+				client.close();
+			}
+		}
+	}
+
+	@Test(groups = {"emulator"}, timeOut = TIMEOUT)
+	public void DocumentWrite() {
```
moderakh, line 71:
```suggestion
	public void documentWrite() {
```

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
```diff
@@ -0,0 +1,254 @@
...
+			fail("Should not throw exception in the test" + ex.getMessage());
+		} finally {
+			if (client != null) {
+				client.close();
+			}
+		}
+	}
+
+	@Test(groups = {"emulator"}, timeOut = TIMEOUT)
+	public void DocumentRead() {
```
moderakh, line 127: please fix the method names as per Java code style.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
```diff
@@ -0,0 +1,254 @@
...
+			fail("Should not throw exception in the test" + ex.getMessage());
+		} finally {
+			if (client != null) {
+				client.close();
+			}
+		}
+	}
+
+	@Test(groups = {"emulator"}, timeOut = TIMEOUT)
+	public void DocumentQuery() {
```
moderakh, line 188: ditto

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/directconnectivity/ClientSideRequestStatisticsTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+public class ClientSideRequestStatisticsTest extends TestSuiteBase {
+
+	@Test(groups = {"emulator"}, timeOut = TIMEOUT)
+	public void AddressResolutionStatistics() {
```
moderakh, line 52: method name, fix as per java method naming.

moderakh (approved):
please fix the minor comments. other than that LGTM.

commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
```diff
@@ -259,17 +261,21 @@ public String toString() {
         private ZonedDateTime startTime;
         private ZonedDateTime endTime;
         private String targetEndpoint;
+        private String errorMessage;
+        private boolean inflightRequest = true;
```
moderakh, line 265: could you please add a comment what this is?

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
kushagraThapar: Good point, I think we can include this because the impact of query plan not being retried on Gateway failed request will be the same. 
We can create a separate PR for this fix, so to run separate CI on just this change, but I think we should fix it as part of the release.

commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
kushagraThapar, line 265: Why do we need this if we can verify this behavior through `endTime` ?
As I recall, if the address resolution is in process, then `endTime` has a value of max time, no ?

On a side note, I think we should initialize `endTime` with null, until we hit `addressResolutionEnd`, thoughts ?

kushagraThapar (changes requested):

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
simplynaveen20: I included in this PR itself,scope is overlapping , as Sams club fail over did happen because of query

commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
simplynaveen20: This is normal our code flow that if one replica gives error we start address refresh in parallel  while talking to other replica, please check code [here](https://github.com/Azure/azure-cosmosdb-java/blob/ca1307963ba53ce4d1aa7bb21447f06c75f10a76/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/StoreReader.java#L296)
So there were several incidences in logs where address call stared but the request got completed before that.
This will give clear information that request is inflight

simplynaveen20: EndTime initialization to null should be fine. Changed it
On InflightRequest, this is normal code flow that if one replica gives error we start address refresh in parallel while talking to other replica, please check code [here](https://github.com/Azure/azure-cosmosdb-java/blob/ca1307963ba53ce4d1aa7bb21447f06c75f10a76/direct-impl/src/main/java/com/microsoft/azure/cosmosdb/internal/directconnectivity/StoreReader.java#L296)
So there were several incidences in logs where address call stared but the request got completed before that.
This will give clear information that request is inflight

simplynaveen20: Added comment

kushagraThapar: Good point. Thanks.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
kushagraThapar: Thanks @simplynaveen20

kushagraThapar (approved):
LGTM

commons/src/main/java/com/microsoft/azure/cosmosdb/ClientSideRequestStatistics.java
```diff
@@ -123,7 +122,7 @@ public String recordAddressResolutionStart(URI targetEndpoint) {
         AddressResolutionStatistics resolutionStatistics = new AddressResolutionStatistics();
         resolutionStatistics.startTime = ZonedDateTime.now(ZoneOffset.UTC);
         //  Very far in the future
-        resolutionStatistics.endTime = ZonedDateTime.of(LocalDateTime.MAX, ZoneOffset.UTC);
+        resolutionStatistics.endTime = null;
```
mbhaskar, line 125: Thanks for this change. It was misleading something to see weird max time.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
```diff
@@ -0,0 +1,252 @@
+/*
```
mbhaskar, line 1: Please change to Microsoft licence

mbhaskar (approved):
Left minor comment regarding license. Please take a look. Thanks for including the QueryPlan change.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/internal/HttpRequestTimeoutTest.java
simplynaveen20: done

PR #387, opened 2021-09-03
Clarify ConnectionPolicy multi write region configs

ealsur:
Current texts imply that write requests will "only" be directed to the first preferred location.

### DIFF
--- a/commons/src/main/java/com/microsoft/azure/cosmosdb/ConnectionPolicy.java
+++ b/commons/src/main/java/com/microsoft/azure/cosmosdb/ConnectionPolicy.java
@@ -315,8 +315,12 @@ public final class ConnectionPolicy {
      * to true has no effect until EnableMultipleWriteLocations in DatabaseAccount
      * is also set to true.
      *
-     * Default value is false indicating that writes are only directed to
-     * first region in PreferredLocations property.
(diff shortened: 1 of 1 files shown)


commons/src/main/java/com/microsoft/azure/cosmosdb/ConnectionPolicy.java
```diff
@@ -277,8 +277,12 @@ public void setEnableEndpointDiscovery(boolean enableEndpointDiscovery) {
      * to true has no effect until EnableMultipleWriteLocations in DatabaseAccount
      * is also set to true.
      *
-     * Default value is false indicating that writes are only directed to
-     * first region in PreferredLocations property.
+     * Default value is false indicating that writes are directed to
+     * first region in PreferredLocations property or the primary account region if no PreferredLocations are specified.
```
simplynaveen20, line 281: first region in PreferredLocations property ->  first region in PreferredLocations property if it's a write region ?

ealsur: In a multi master account they are all write regions?

simplynaveen20 (approved):
Thanks Matias

PR #392, opened 2021-10-18
Adding group id explicitly to poms for partner release pipeline validation

simplynaveen20:
Adding group id explicitly to poms for partner release pipeline validation


benchmark/pom.xml
```diff
@@ -31,7 +31,7 @@
     <artifactId>azure-cosmosdb-parent</artifactId>
     <version>2.6.14</version>
   </parent>
-  
+  <groupId>com.microsoft.azure</groupId>
```
weshaggard, line 34: You should add the version property as well.

PR #402, opened 2022-02-28
Fixing the response header 'x-ms-continuation' value issue.

aayush3011:
1. The issue is happening when using the top query to get the top of 'x'  items where 'x' is greater than the total number of items in the database.
2. Link for the ICM - https://portal.microsofticm.com/imp/v3/incidents/details/286671765/home

### DIFF
--- a/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/TopDocumentQueryExecutionContext.java
+++ b/gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/TopDocumentQueryExecutionContext.java
@@ -127,9 +127,14 @@ public class TopDocumentQueryExecutionContext<T extends Resource> implements IDo
                     if (top != collectedItems) {
                         // Add Take Continuation Token
                         String sourceContinuationToken = t.getResponseContinuation();
-                        TakeContinuationToken takeContinuationToken = new TakeContinuationToken(top - collectedItems,
-                                sourceContinuationToken);
-                        headers.put(HttpConstants.HttpHeaders.CONTINUATION, takeContinuationToken.toJson());
+                        if (sourceContinuationToken != null) {
+                            TakeContinuationToken takeContinuationToken = new TakeContinuationToken(top - collectedItems,
+                                    sourceContinuationToken);
+                            headers.put(HttpConstants.HttpHeaders.CONTINUATION, takeContinuationToken.toJson());
+                        } else {
+                            // Null out the continuation token
+                            headers.put(HttpConstants.HttpHeaders.CONTINUATION, null);
+                        }
                     } else {
                         // Null out the continuation token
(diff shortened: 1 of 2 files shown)


gateway/src/main/java/com/microsoft/azure/cosmosdb/rx/internal/query/TopDocumentQueryExecutionContext.java
```diff
@@ -127,9 +127,14 @@ public Boolean call(FeedResponse<T> frp) {
                     if (top != collectedItems) {
                         // Add Take Continuation Token
                         String sourceContinuationToken = t.getResponseContinuation();
-                        TakeContinuationToken takeContinuationToken = new TakeContinuationToken(top - collectedItems,
-                                sourceContinuationToken);
-                        headers.put(HttpConstants.HttpHeaders.CONTINUATION, takeContinuationToken.toJson());
+                        if (sourceContinuationToken != null) {
```
xinlian12, line 130: 1. Can we add a unit test for this fix?
2. Can we also add a comment here to explain why we can use sourceContinuationToken can be used as a signal to determine it is the last page?
3. Will this change be ported to V4? since V4 also has the same issue?

aayush3011: 1. Added the unit test case
2. Added the comment
3. Yes, I am working on the change for V4 as well. Will send out the PR for that as well soon.

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/TopQueryTests.java
```diff
@@ -143,6 +143,12 @@ public void queryDocumentsWithTopContinuationTokens() throws Exception {
         this.queryWithContinuationTokensAndPageSizes(query, new int[] { 1, 5, 10 }, 8);
     }
 
+    @Test(groups = { "simple" }, timeOut = TIMEOUT * 1000, retryAnalyzer = RetryAnalyzer.class)
+    public void queryDocumentsWithTopGreaterThanItemsContinuationTokens() throws Exception {
+        String query = "SELECT TOP 40 * FROM c";
```
xinlian12, line 148: How do we know 40 is more than what we have in the container?  
maybe we can use Integer.max or utilize the doc.size to guarantee the value we choose here is definitely higher what we have created in the container?

xinlian12 (approved):
LGTM, thanks for the fix

sdk/src/test/java/com/microsoft/azure/cosmosdb/rx/TopQueryTests.java
aayush3011: Using Integer.max

kushagraThapar (approved):
LGTM, thanks @aayush3011

